PR #7, opened 2016-08-07
Make sure there are no conflicts in the MAME OPL interface

kazblox:
For the time being, MAME OPL support is nulled out due to heavy sound changes but to make sure, the references to the dbopl header files in the interface are removed to prevent major conflict in compilation if it gets added back in.


kazblox:
closed because this leftover is likely getting removed.

PR #49, opened 2016-12-14
Merge RIVA 128 and RIVA TNT implementations

MoochMcGee:
This should allow for more nVidia card implementations with less work and less maintenance cost.


MoochMcGee:
In fact, this should work until G80, which released in 2006. Only then would we need a new implementation.

MoochMcGee:
Whoops, hold on, I need to fix some bugs.

MoochMcGee:
K, that should be good, it compiles now.

PR #78, opened 2017-01-15
Update sound_sb.c

basic2004:
I solved OPL3 problem on SB Pro V2, SB16, AWE32.
This was defined 388h-389h only, so plays like OPL2.
OPL3 will plays correctly if this defined 388h-38bh.


OBattler:
Thank you very much for your pull request! I merged it now.

PR #79, opened 2017-01-15
CMS/Game Blaster pitch correction

basic2004:
This patch make more accurate CMS emulation like actual GameBlaster.
Original works of this patch from https://www.vogons.org/viewtopic.php?f=41&t=38350 for DOSBox, thanks to them.

### DIFF
--- a/src/sound_cms.c
+++ b/src/sound_cms.c
@@ -7,6 +7,8 @@
 #include "sound.h"
 #include "sound_cms.h"
 
+#define MASTER_CLOCK 7159090
+
 typedef struct cms_t
 {
         int addrs[2];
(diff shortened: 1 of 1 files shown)


basic2004:
for example

https://www.youtube.com/watch?v=hF6YpwQ8YEk : Actual Sound Blaster 2.0 CMS

https://www.youtube.com/watch?v=SEyIcNDDAB8 : DOSBox recording, 86Box plays like this raised pitch too.

OBattler:
Oops... I didn't notice this and manually carried over the same patch of yours from mainline PCem. Will review this later to see if there's any differences.

PR #90, opened 2017-01-26
Change ATI 18800 BIOS

MoochMcGee:


MoochMcGee:
The BIOS used here is available here: http://vgamuseum.info/index.php/cards/item/68-ati-18800-1-ega-vga-wonder

PR #105, opened 2017-04-02
Fix a lil MMIO bug in RIVA 128.

MoochMcGee:


MoochMcGee:
This doesn't crash anymore.

PR #132, opened 2017-06-07
Update README.md

dhrdlicka:
This definitely needs extending with sections on building, using and similar, but at least we now have a decent starting point instead of the current provisional readme.

Discuss. Also I would prefer not merging this PR in until this is actually complete.


MoochMcGee:
@dhrdlicka This PR looks REALLY messy. Perhaps it's time to close it?

dhrdlicka:
Would be a good idea, really.

PR #144, opened 2017-07-12
Some D3D fixes, and monochrome monitor...

basic2004:
This fix will no more crash silently when locking Windows with Direct3D renderer.

Known issues
1. When locking Windows 10, stuck screen with windowed mode.
solution : Change renderer to DirectDraw and change to Direct3D.
2. When locking Windows 10, showing fatal error with fullscreen mode.
this error instead crash.
3. Crashes when ran 8088mph with composite CGA, D3D renderer, fullscreen mode.
window mode is solved but fullscreen mode isn't yet.

And I patched monochrome monitor palette correctly when entering and leaving fullscreen and changing renderer.


OBattler:
Pull request merged, thanks!

PR #174, opened 2017-08-27
Lowercase all CAPS to prevent case sensitive errors in non-Windows platforms

claunia:


waltje:
Merged.  Without a rage, even!

waltje:
Merged.

PR #200, opened 2017-10-09
No need to have extensions everywhere in makefile, this just difficul…

claunia:
…ts cross compiling


waltje:
Partly implemented.

PR #281, opened 2018-04-11
 Updated Nuked OPL3 1.7.4 to 1.8

basic2004:
Test this before commit.

from Nukeykt:

Nuked OPL3 v1.8 is out.
Changelog:
* New envelope generator.
* Rhythm mode emulation is 100% correct now.

These and other aspects of chip have been verified on YMF262 die shot, so emulation should be perfect now.


OBattler:
I have something similar already done on my end but I'll look at this too and see if there's any differences and what they affect.

OBattler:
OK, seems this lacks the required changes to snd_dbopl.cc to use WriteRegBuffered. I'll see if there's other differences from my own port.

basic2004:
Yeah, I checked updates, that contained latest Nuked OPL3 and works fine.

Issue #381, opened 2019-04-27
[BUG] Resizable Window issues


ppgrainbow:
I would like to report a couple things when it comes to making a window in 86Box resizable.

1. When I make the 86Box window resizable (select View > Resizable Window) from fixed, the menu options (Action, View, Tools and Help) appear to be overlapped by the window itself, thus messing up the menu.

This happens upon start up as well. Here's a screenshot for clarification:

![bug1](https://user-images.githubusercontent.com/32658796/56856314-b6dd6f00-690c-11e9-92d8-1a2a939060fb.png)

2. When the Resizable Window feature is enabled (selected under View > Resizable Windows), when I try to resize the 86Box window, 86Box crashes. Here's a screenshot for proof:

![bug2](https://user-images.githubusercontent.com/32658796/56856317-c3fa5e00-690c-11e9-98bf-c37b07c09262.png)

The crashing bug can only be produced when I have set the render to SDL and that I resize the window. I found no problems with resizing the window when the DirectDraw and Direct3D 9 render is enabled.

Please consider fixing the bugs that I provided. Thanks. :)

Until the bug is fixed, I would avoid using the SDL render to resize the window.

ppgrainbow:
The SDL render resize bug has been fixed and I'm waiting for it to be ported to the Jenkins build. Until then, I'll leave this bug opne.

driver1998:
Issue 1 is related to https://github.com/86Box/86Box/commit/2da25c0a29aa574a2d1ea752c6ef844758de5404

This is for fixing a window-border related bug in clang/llvm-mingw (and mingw 6.0 I guess), where the mouse cursor can get out of the window when it is locked.
But it does cause the problem you mentioned.

PR #398, opened 2019-07-27
win_ui: Proper fix to window border issues

driver1998:
Who know calculating window border sizes can be this hard, it can even be affected by compiler and linker settings...

This fixes https://github.com/86Box/86Box/issues/381 (caused by https://github.com/86Box/86Box/commit/2da25c0a29aa574a2d1ea752c6ef844758de5404).

### DIFF
--- a/src/win/win_ui.c
+++ b/src/win/win_ui.c
@@ -8,7 +8,7 @@
  *
  *		user Interface module for WinAPI on Windows.
  *
- * Version:	@(#)win_ui.c	1.0.39	2019/3/20
+ * Version:	@(#)win_ui.c	1.0.40	2019/7/28
  *
  * Authors:	Sarah Walker, <http://pcem-emulator.co.uk/>
  *		Miran Grca, <mgrca8@gmail.com>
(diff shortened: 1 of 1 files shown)


src/win/win_ui.c
```diff
@@ -349,15 +348,22 @@ MainWindowProcedure(HWND hwnd, UINT message, WPARAM wParam, LPARAM lParam)
...
-
 				/* Main Window. */
-				MoveWindow(hwnd, rect.left, rect.top,
-					unscaled_size_x + non_client_width,
-					unscaled_size_y + non_client_height + sbar_height,
-					TRUE);
+				if (GetSystemMetrics(SM_CXPADDEDBORDER) == 0) {
+					/* For platforms that subsystem version < 6.0 (default on mingw/msys2) */
+					/* In this case, border sizes are different between resizable and non-resizable window */
+					MoveWindow(hwnd, rect.left, rect.top,
```
driver1998: If Windows XP support is not considered, we can just force the subsystem version to 6.0 and get rid of this nightmare...

OBattler: I merged your pull request now, but yes, I do think we should force the subsystem version to 6.0, as I don't really think supporting Windows XP makes sense anymore.

PR #466, opened 2019-11-19
Added a SDL (Hardware) renderer mode distinct from SDL (Software)

Agetian:
A simple patch that differentiates between SDL (Software) and SDL (Hardware) render modes. Can't test how SDL works on Windows since I don't have it, but a TODO comment says that it seems to act weird on Windows 7 in hardware mode, disabling Aero and not playing ball with fullscreen mode. Not sure if that's still the case and if it holds true on Win 8/10, but at least on Linux, it doesn't have any side effects but looks much better than Software since scale quality can be improved (and it's the best option on Wine that interacts the best with full screen mode, other modes tend to be funky when exiting full screen). From hence, two modes, one to launch SDL Software renderer and the other to launch SDL Hardware renderer with a scale quality hint. Might also be helpful on Windows if it doesn't cause the side effect outlined above anymore.

Things to establish:
Does SDL (Hardware) still act funny on Windows 7?
Does SDL (Hardware) act wrong/weird on newer Windows (8/8.1/10)?

**Potential issue mentioned by OBattler:**
It seems that on Windows, running SDL in Hardware mode may cause the screen to render black or to stop updating once the guest resolution changes. Unable to reproduce this so far on my Linux Wine setup with SDL v2.0.10, will try a Win10 VM as soon as I have some time.

I tried my best to format this consistently with the style used, but I can't for the life of me figure out what's wrong with the tabs formatting in resource.h for IDM entries (SDL_SW and SDL_HW). I tried two tabs, three tabs, expanded tabs, unexpanded tabs in Vim, but nothing seems to match exactly :/ 

Testing is welcome on Windows machines since I don't have one.

In case SDL (Hardware) doesn't really have any side effects on Windows anymore (e.g. on the current SDL 2 revision), maybe it's worth just defaulting to a hardware renderer with a scale hint since it looks a lot nicer and in that case this patch would be unnecessary at all (however, maybe it would be wise to keep at least one Software renderer in the list? :/ ). If there is an unwanted side effect, this split may be nice in case Hardware works better for someone (on e.g. an unaffected version of Windows or on Linux/OS X running via Wine), but Software is still available in case HW mode causes issues but other renderers are unavailable or act up on the given platform for any reason.


MoochMcGee:
Wait a second, how did you run 86Box on Linux? o.o 86Box doesn't work on Linux except through Wine as far as we know. We'd be very happy to see another PR if you managed to get a native Linux build running though!

Agetian:
Yeah, I'm cross-compiling it on Linux using mingw32, then running the resulting build via Wine. It works pretty much flawlessly that way (tested on Wine 4.4-staging), the only thing I noticed is that D3D9 / DDraw would sometimes "hang" on an attempt to leave fullscreen, thus SDL (Hardware) would be the preferred option for that, offering the same quality as the other two renderers and not having the side effect mentioned. Also, this option might actually work well on Windows too (offering better quality in SDL renderer mode) if it doesn't cause the side effect mentioned.

Would also love to see a native Linux build at some point, but I understand that it might be a complex undertaking, so I'm happy to use Wine for the time being to cross-compile and run it. Sadly, I'm not experienced at making UIs, so I don't think I'll be the best person to contribute a full working native Linux platform code with the UI and stuff. That said, I'd be more than happy to contribute in any way I can, including testing on various distros and in different modes / with different renderers, different UI toolkits (wx, GTK, Qt, etc., not sure what would be preferred for a Linux version tbh; I think wx might be nice for dual *nix/Windows compatibility).

MoochMcGee:
Which Linux distro are you using? I've tried cross-compiling 86box with Arch, but I haven't really been able to get it to work, sadly :c

Agetian:
I'm on Linux Mint 18.3, I had to tweak the Makefile to get it to work, it's actually somewhat hacky, I just edited it really quickly just to make it compile. I had to cross-compile my own OpenAL, SDL2, Freetype, etc., of course, before I could get the main compile to run. I can post my Makefile if you want :) Basically, I just do "make -j8 -f win/Makefile.mingw32-cross" to compile and it works (if all the dependencies are satisfied and all the prerequisite libs are compiled), mingw32-cross is my "changed" version of the Makefile specifically for cross-compilation

MoochMcGee:
BTW, do you think you could maybe join #86box on Freenode? I'm currently banned from the IRC network the main 86box IRC channel is on due to something stupid I did several months ago while I was off my pills.

Agetian:
Don't mind at all, and I'm on 86box Discord too (joined today), will need a bit of time to get IRC up and running :) Will try later today

MoochMcGee:
IRC is uh, pretty simple, you just need an IRC client (I recommend Hexchat) and to join a network, and then type in "/join #86box" no quotes and hit enter.

Agetian:
I'll try working on reproducing and fixing the black screen issue on Windows soon (will need to set up a VM for it), will keep this PR updated and will commit a fix if I devise one.

Agetian:
Will close for now, working on some restructuring / improvements and need to resubmit the PR anyway. Will reopen once ready :)

PR #468, opened 2019-11-20
WIP: Implement SDL (Hardware) render mode separate from SDL (Software)

Agetian:
This is a revised PR for implementing a SDL (Hardware) render mode distinct from SDL (Software), it fixes an issue with the black screen render on Windows systems.

A simple patch that differentiates between SDL (Software) and SDL (Hardware) render modes. When ruinning on Linux via Wine, SDL (Hardware) doesn't have any side effects but looks much better than Software since scale quality can be improved (and it's the best option on Wine that interacts the best with full screen mode, other modes tend to be funky when exiting full screen). On Windows, it offers superior scaling quality compared to SDL (Software) in case SDL rendering is used. From hence, two modes, one to launch SDL Software renderer and the other to launch SDL Hardware renderer with a scale quality hint. 

Tested on Win 10, Win 8 (VM), Linux Mint 18.3 via Wine 4.4-staging.

A TODO line mentioned in win_sdl.c that running SDL2 in Hardware mode could cause some side effects on Windows such as turning off Aero. Quite possibly, this PR fixes that (but additional testing is welcome). In case SDL (Hardware) doesn't really have any side effects on Windows anymore (e.g. on the current SDL 2 revision), good, we have a hardware-accelerated SDL2 mode with good scaling. If there is an unwanted side effect, this split may be nice in case Hardware works better for someone (on e.g. an unaffected version of Windows or on Linux/OS X running via Wine), but Software is still available in case HW mode causes issues but other renderers are unavailable or act up on the given platform for any reason.


dhrdlicka:
Merged in 26c0a5d2eaaf3a87f48ecb34ca862e4bf1e88d34

PR #496, opened 2019-12-14
Adding the IBM Ambra's on 86Box

tiseno100:
The IBM Ambra DP60 and DP90 use the standard Intel OEM boards but with their own special BIOS. These machines weren't that famous but they can be a good and special implementation for 86Box.


tiseno100:
Contact me for the roms on Discord. <coolgamerxd>

PR #579, opened 2020-01-24
Machine table fixes

nerd73:
- Removed VLB from all PCI boards as none that are emulated have VLB slots
- Changed max ram of the PB410A and LS486E to 36 MB and 128 MB respectively.
- ISA/VLB boards now identified in board name
- 3.3V only Socket 7 boards now grouped together
- Some spacing fixes


ppgrainbow:
nerd73, since the PB410A has 4 MB of RAM soldered onto the motherboard, should the amount of memory installed between 4 MB and 36 MB be in 4 MB increments and not 1 MB increments?

PR #592, opened 2020-02-04
Rawinput Joystick, and some other minor fixes

driver1998:
RawInput joystick:
- Supports 32 buttons, 6 axis and 4 POVs per joystick (in theory)
- Supports 8 Joysticks (also in theory)
- Tested with Xbox 360 controller, and two more generic HID compatible controllers
- DirectInput is used by default, RawInput enabled when `DINPUT=n` is specified in makefile.

Fixes:
- Add proper declaration and lib of SetCurrentProcessExplicitAppUserModelID.
- Fix the progress bar in new disk image dialog
- Now all RawInput data (keyboard, mouse, joystick) are processed at once.

### DIFF
--- a/src/printer/prt_escp.c
+++ b/src/printer/prt_escp.c
@@ -2006,7 +2006,7 @@ read_ctrl(void *priv)
 {
     escp_t *dev = (escp_t *)priv;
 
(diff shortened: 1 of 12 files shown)


driver1998:
Xinput support is removed from makefile for now for those who wonders.

PR #599, opened 2020-02-17
Added the Aztech Sound Galaxy Washington/Clinton series of cards.

TC1995:
Added the Aztech Sound Blaster DSP clones of DSP types.
Fixed OPL enable/disable in WSS.
Added IRQ selection and Addr selection functions to the MPU-401 code.

### DIFF
--- a/src/sound/snd_azt2316a.c
+++ b/src/sound/snd_azt2316a.c
@@ -1,0 +1,1340 @@
+/*
+ * TYPE 0x11: (Washington)
+ * Aztech MMPRO16AB,
+ * Aztech Sound Galaxy Pro 16 AB
+ * Aztech Sound Galaxy Washington 16
+ * ...and other OEM names
+ * FCC ID I38-MMSN824 and others
+ *
+ * TYPE 0x0C: (Clinton)
+ * Packard Bell FORTE16
+ * Aztech Sound Galaxy Nova 16 Extra
+ * Aztech Sound Galaxy Clinton 16
+ * ...and other OEM names
+ *
+ * Also works more or less for drivers of other models with the same chipsets.
+ *
+ * Copyright (c) 2020 Eluan Costa Miranda <eluancm@gmail.com> All rights reserved.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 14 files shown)


TheVirtualComputer:
Why is this (apparently) licensed under the GPL v3? PCem and 86Box are both licensed under GPL v2, and I am not at all sure if all of the copyright holders for the PCem and 86Box code have even given the necessary permission to use a later version of the GPL, so if this were to be used with code that was made available under the GPL v2 without the upgrade clause then this would be a clear license violation.

So I would suggest for the copyright holder of this code (EluanCM) to allow use of it under the GPL v2 in addition to the GPL v3, so that it would be legally usable in both emulators.

I was also planning on adding emulation of Aztech and other Crystal-based sound cards to my own Virtual Computer (which will be seeing a new release soon), but this obvious license conflict would have to be resolved first before I would be able to release it.

TC1995:
Where do you see GPL v3?

TheVirtualComputer:
The GPL v3 clause can very clearly be seen in the snd_azt2316.c file, where it clearly states the following:
***
* Copyright (c) 2020 Eluan Costa Miranda <eluancm@gmail.com> All rights reserved.
*
* This program is free software: you can redistribute it and/or modify
* it under the terms of the GNU General Public License as published by
* the Free Software Foundation, either _version 3_ of the License, or
* (at your option) any later version.
*
* This program is distributed in the hope that it will be useful,
* but WITHOUT ANY WARRANTY; without even the implied warranty of
* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
* GNU General Public License for more details.
*
* You should have received a copy of the GNU General Public License
* along with this program.  If not, see <http://www.gnu.org/licenses/>.**

Either the "version 3" is a typo (mistyping of "version 2") or the code is not properly licensed.

TC1995:
Well, the license file of 86box called "COPYING" says GPL v2 or later.

src/sound/snd_azt2316a.c
```diff
@@ -0,0 +1,1340 @@
...
+ * ...and other OEM names
+ *
+ * Also works more or less for drivers of other models with the same chipsets.
+ *
+ * Copyright (c) 2020 Eluan Costa Miranda <eluancm@gmail.com> All rights reserved.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
```
dhrdlicka: This code will have to be relicensed to GPLv2.

PR #614, opened 2020-03-21
Fix build with new dynarec disabled

richardg867:


richardg867:
A keyboard controller fix got accidentally rolled into this PR as well.

PR #616, opened 2020-03-21
PGA370 boards bringup

tiseno100:
Capture

A Socket 370 440BX board. The addition is mostly meant to fit the future VIA C3 chips added by nerd. There are surely more to come if the PR gets accepted.


qeeg:
Hey, I'm pretty sure that for 440ZX boards, we'd prefer that the chipset were to be emulated before the board is added, so as to ensure that the board functions properly.

tiseno100:
The 440ZX is just an underpowered 440BX that runs at 66Mhz only and has a 256MB of RAM limit. Meant for cheap Celeron computers. They have zero difference in terms of registers or vendor IDs. That's why most drivers report 440BX/ZX.

qeeg:
Maybe you should add a comment in m_at_socket8.c to reflect this then?

PR #619, opened 2020-03-24
Separate PIIX4 and PIIX4E

richardg867:
Windows does not make the distinction but some tools do. Will trigger redetection of PCI devices.


richardg867:
Already done, no need for this PR anymore.

PR #624, opened 2020-03-25
Added the Aztech Sound Galaxy 16 cards.

TC1995:
Added the reset function of MCA from PCem.
Made the MCA devices use the reset parameter if available.


qeeg:
The code still needs to be relicensed...

TC1995:
Well, from what I've been told by Eluan, he said he doesn't care about the license, be it GPLv2 or v3.

qeeg:
Then change the licensing text in your code :p

PR #645, opened 2020-03-26
Added PC Partner MB520N

nerd73:
The PC Partner MB520N is a Socket 7 430VX board with AMI WinBIOS, using SMC 669 Super I/O and SST 29EE010 flash.


nerd73:
Let this stand as a lesson to you all: Just because it POSTs doesn't mean it works.

nerd73:
65

PR #682, opened 2020-04-10
Covington Celeron and Katmai Pentium III emulations

anabate123:


tiseno100:
Umm, we don't emulate SSE yet for the Pentium IIIs

PR #702, opened 2020-04-18
Added the requires VNC libraries in documentation + missing instructions

tiseno100:


dhrdlicka:
> We use proprietary VNC libraries so make sure to replace those with the *vnclibs.zip* contained in the documentation.
I am quite reluctant to distributing libraries with questionable licenses as a part of the documentation, so we should either just get it working with whatever MSYS ships with, or dynaload it.

dhrdlicka:
By the way, your master branch is broken. :P You might want to delete and recreate your fork.

PR #707, opened 2020-04-20
Dev-branch machines with Super I/O chips that will not be implemented in v2.10

nerd73:
Namely, the AMI Apollo and the Iwill P55XB2. Also, while it's Super I/O chip is implemented, the Packard Bell Bora Pro is being dev-branched as well for failing to POST.

Also adds a 66 MHz FSB Pentium II table in preparations for 440LX machines to be added in the future.


tiseno100:
The Bora Pro was always meant to be dev branched. I dunno who removed it off. Nice btw.

Asternmerr642:
> 
> 
> The Bora Pro was always meant to be dev branched. I dunno who removed it off. Nice btw.
> 
> You forgot doing the QDI Advance 2, Gigabyte 6BXC & i think some others who need the IT8671 or 8661F

I agree.

PR #709, opened 2020-04-20
Dynamically load Win7+ APIs

driver1998:
This makes it possible to build 86Box to run on Windows XP again.

Note:
If compile on MSYS2, you'll need to [downgrade winpthread to a version that supports XP](https://github.com/msys2/MINGW-packages/issues/5139#issuecomment-510575800).
And then dynamically link OpenAL (and dynamically links to the downgrade winpthread DLL), or build OpenAL again and use static library.

### DIFF
--- a/src/win/win.c
+++ b/src/win/win.c
@@ -19,7 +19,6 @@
  *		Copyright 2017-2019 Fred N. van Kempen.
  */
 #define UNICODE
-#define NTDDI_VERSION 0x06010000
 #include <windows.h>
 #include <shlobj.h>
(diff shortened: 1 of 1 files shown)


dhrdlicka:
I am not sure if there is any point in maintaining Windows XP support tbh.

driver1998:
Well that does seem rather pointless... Just to show that it is easy to do though.

dhrdlicka:
It's used to group emulator and manager windows under a single taskbar icon in Win7+

PR #710, opened 2020-04-21
Added more speeds to VIA Cyrix III (Samuel) + clock multipliers

anabate123:
What would it be like without the remaining speeds to Cyrix III Samuel?


anabate123:
Sorry I changed the clock speeds on the Celeron Mendocino, I corrected it by changing it back.

anabate123:
Nobody responded to it.

PR #739, opened 2020-05-02
Added various machines

amdk6:
**Fujitsu ICL ErgoPro x453**: Machine somewhat works, but requires a PC87308 Super I/O chip and an apparent PSU controller.
**Acer V60N**: Machine almost works, but mysteriously hangs at 5F.


tiseno100:
try using the AMI keyboard controller. It might "resolve" keyboard errors. Also the PC87308 with the PC87307 aren't that different. So it's possibility to stay is medium to high.

amdk6:
> 
> 
> try using the AMI keyboard controller. It might "resolve" keyboard errors. Also the PC87308 with the PC87307 aren't that different. So it's possibility to stay is medium to high.

will do.

amdk6:
Gotta close this pull request, as my fork was a mess.

PR #797, opened 2020-06-10
Added the Gateway 2000 OEM-based Zappa ROM

anabate123:
I want to see if I could add the Gateway 2000 OEM-based Zappa ROM.


src/include/86box/machine.h
```diff
@@ -270,6 +270,7 @@ extern int	machine_at_430nx_init(const machine_t *);
 extern int	machine_at_p54tp4xe_init(const machine_t *);
 extern int	machine_at_endeavor_init(const machine_t *);
 extern int	machine_at_zappa_init(const machine_t *);
+extern int      machine_at_gw2kzp_init(const machine_t *);
```
dhrdlicka, line 273: Fix the indenting here, please

anabate123: I fixed the identing.

PR #805, opened 2020-06-12
Added the Gateway 2000 Aries (Non-Integrated) ROM

anabate123:
I want to see if I can add the Gateway 2000 Aries (Non-Integrated ROM)


anabate123:
It failed, so this is closed.

PR #944, opened 2020-07-24
Add Virtual PC 2007 machine

richardg867:
Not working yet, needs the `vpcext` (0f 3f) opcode among other things.


ab310714gz4700:
Virtual PC 2004 machine is missing

PR #951, opened 2020-07-28
PostScript printer: actually save .ps files if GhostScript is absent

richardg867:
The message box and documentation say .ps files are saved if GhostScript is absent, but the current code only leaves the temporary .tmp files behind. Fix that.


dhrdlicka:
It might be better to rename them to `.ps` only at the point when Ghostscript would have been invoked, so you can tell apart unfinished and finished documents.

PR #1001, opened 2020-09-12
Some final touches on the Intel 82335

tiseno100:
Most I could potentially implement are now complete
- Added some commentary
- switched some complex algorithms into definitions for the sake of the code being clean
- Implemented the ROM size determination register just for some Shadow RAM sanity


src/chipset/intel_82335.c
```diff
@@ -31,16 +31,25 @@
...
+/* R/W operator for the Video RAM region */
 #define determine_video_ram_write_access ((dev->regs[0x22] & (0x08 << 8)) ? rw_shadow : ro_shadow)
 
+/* Base System 512/640KB switch */
 #define enable_top_128kb (MEM_READ_INTERNAL | MEM_WRITE_INTERNAL)
 #define disable_top_128kb (MEM_READ_DISABLED | MEM_WRITE_DISABLED)
 
+/* ROM size determination */
+#define rom_size ((dev->regs[0x22] & (0x01 << 8)) ? 0xe0000 : 0xf0000)
+
```
dhrdlicka, line 52: It's common practice to name #define's using ALL CAPS.

dhrdlicka (changes requested):

src/chipset/intel_82335.c
tiseno100: I did it mostly in sake of clarity. Even if they do the same function. It's the first time I use that strategy.

Also all defines have been capped now on the new commit below

PR #1017, opened 2020-10-04
Added Gateway 2000 Tigereye and Venus ROMs

anabate123:
Gateway 2000 Tigereye (features MMX)
image

Gateway 2000 Venus
image


tiseno100 (changes requested):
Please use the naming Gateway gave to those machines without their clock speed present.

PR #1026, opened 2020-10-12
Fix PostScript printer

richardg867:
The PostScript printer code was ending the document upon reaching 64k of PostScript data, instead of just flushing to the temporary file.


dhrdlicka:
Seems like piix4 stuff sneaked in there.

dhrdlicka:
Nevermind, seems to be just indenting

PR #1067, opened 2020-11-12
Small Makefile fixes

lemondrops:
Summary
=======
The Makefile silences the command output in some cases by redirecting it to NUL, which, when run under `bash` in an MSYS2 environment results in the output being redirected to a regular file called `NUL`, which then can't be deleted by standard Windows methods due to NUL being a reserved name.

Approach
========
Since the Makefile is obviously not intended to be run under `cmd.exe`, the input is now redirected to `dev/null` instead. Additionally, an unused leftover variable has been removed as well.

Checklist
=========
* [ ] Closes issue #xxx
* [ ] This pull request requires inclusion of a ROM in the romset
* [x] My commit messages are descriptive and I have not added any irrelevant files to the repository

### DIFF
--- a/src/win/Makefile.mingw
+++ b/src/win/Makefile.mingw
@@ -66,9 +66,6 @@ ifeq ($(DEV_BUILD), y)
  ifndef PAS16
   PAS16		:= n
  endif
- ifndef PORTABLE3
-  PORTABLE3	:= y
- endif
  ifndef PS1M2133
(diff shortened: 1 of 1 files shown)


dgurney:
Nitpick: try to be more succinct with your commit message subject lines next time to avoid truncation issues like this:
image
Ideally the subject line should be no longer than 50 characters, any extra details should go to the body.

PR #1098, opened 2020-11-18
Add VHD Support

### LINKED ISSUE #731: Adding support of dynamically expanding disk images
^
Can we have dynamically expanding disk images in any format? (VHD images at least)
We can save space with these zeros gone (you can do that too with NTFS compression, but dynamic images is nice.)

sards3:
Summary
=======
This PR adds full support for VHDs. Fixed, dynamic, and differencing VHDs are supported. Support was added to the GUI for these options.  VHD support was implemented using [MiniVHD](https://github.com/shermp/MiniVHD).  The previous VHD code, which only supported fixed VHDs, was removed.

**Note: This seems to work, but I have not done extensive testing. Additional testing by the community is recommended.**

Checklist
=========
* [x] Closes #731


shermp:
@sards3 you may want to include the current round of fseek/ftell changes I've just pushed to MiniVHD. It might actually work for everyone now :) (awaiting confirmation of those using classic MinGW that it works)

sards3:
@shermp Done.

PR #1150, opened 2020-12-05
Don't warn on machine reset if in standalone settings mode

lemondrops:
Summary
=======
When 86Box is run in settings-only mode (with the `-S` switch), it no longer warns about a hard reset of a machine when saving the machine configuration since there's no machine running.

Checklist
=========
* [ ] Closes #xxx


dhrdlicka:
IMHO that warning doesn't need to be displayed at all in standalone settings mode.

PR #1187, opened 2020-12-22
Fix blit thread related deadlocks

luisjoseromero:
Summary
=======
Removed unnecessary calls to video_wait_for_blit(), which in
addition to not doing much were also in a race with the
video_blit_memtoscreen() thread to wait for and reset the
blit_data.blit_complete event signalled by blit_thread(),
causing several deadlocks.

### DIFF
--- a/src/win/win.c
+++ b/src/win/win.c
@@ -719,7 +719,6 @@ plat_setvid(int api)
 
     win_log("Initializing VIDAPI: api=%d\n", api);
     startblit();
-    video_wait_for_blit();
 
     /* Close the (old) API. */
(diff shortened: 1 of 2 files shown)


src/win/win.c
```diff
@@ -764,7 +762,6 @@ plat_vidapi_enable(int enable)
     if (!vid_api_inited || !vid_apis[vid_api].enable)
 	return;
 
-    video_wait_for_blit();
```
luisjoseromero, line 767: This was the most insidious one, because it was the once called during WM_SIZE, and thus the easiest one to reproduce the issue.

PR #1194, opened 2020-12-23
Added NCR and Olivetti AT-class machines:

EngiNerd89:
Summary
=======
Added the following new machines:
- NCR PC8/810/710/3390/3392
- NCR 3302
- Olivetti M290

Checklist
=========
* [ ] Closes #xxx
  * [X] I have opened a roms pull request - https://github.com/86Box/roms/pull/93

References
==========
NCR PC-8/810/710/3390/3392
Note: several 286-class PCs were produced by NCR with different model names, depending on the case form-factor and on the mainboard layout (passive backplane with ISA CPU card or active mainboard). However, the chipset being used in all those models is the same, and BIOS chips appear to be interchangeable.
- Press releases: http://www.thecorememory.com/html/ncr_servers___pcs.html
- Manual: http://www.thecorememory.com/NCR_286_TRM.pdf

NCR 3302
Note: Chips & Technologies NEAT chipset (82C206, 82C211, 82C212, 82C215), PVGA1A VGA card on-board, IDE controller onboard
- Info on NCR 3000 Series PCs: https://www.ardent-tool.com/NCR/Docs/NCR_3000_Series.pdf
- Mainboard jumper settings: https://stason.org/TULARC/pc/motherboards/N/NCR-CORPORATION-286-CLASS-3302-80286-TYPE-I.html
- Manual unavailable :-(

Olivetti M290:
Note: several Olivetti models are identified by the name M290-XX. They have different mainboards with different chipsets and BIOSes. The machine added here is the first M290 (without dash and digits).
- Service guide: https://www.ardent-tool.com/Olivetti/Docs/service_guide/systems1/cap5.pdf
- Reference disk (required for BIOS setup): https://web.archive.org/web/20150608124240/http://www.retrocomputing.net/parts/olivetti/m290/docs/m290ct.exe

### DIFF
--- a/src/device/keyboard_at.c
+++ b/src/device/keyboard_at.c
@@ -13,10 +13,12 @@
  * Authors:	Sarah Walker, <http://pcem-emulator.co.uk/>
  *		Miran Grca, <mgrca8@gmail.com>
  *		Fred N. van Kempen, <decwiz@yahoo.com>
+ *      EngiNerd <webmaster.crrc@yahoo.it>
  *
  *		Copyright 2008-2020 Sarah Walker.
  *		Copyright 2016-2020 Miran Grca.
  *		Copyright 2017-2020 Fred N. van Kempen.
+ *      Copyright 2020 EngiNerd.
  */
 #include <stdio.h>
 #include <stdint.h>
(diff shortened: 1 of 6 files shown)


tiseno100:
It'll be nice to have things like

static uint8_t
m290_read(uint16_t port, void *priv)

on a split device which you can add on the machine like how I did the Phoenix 486 jumpers for the sake of clarity.

EngiNerd89:
Thanks for the remark. I'll fix that for the next pull request. To this aim, which of the following strategies would you recommend: one c file per device, or one c file per manufacturer (i.e., all jumper setting devices for that manufacturer in the same file)?

PR #1359, opened 2021-04-06
Full screen resize fix

### LINKED ISSUE #1355: Video misaligned after full screen
**Describe the bug**
After returning from full screen mode, video output is misaligned and not scaled correctly.
The bug seems to have been introduced in build 2810 (regular). Previous build works correctly.

**To Reproduce**
1. Start a machine
2. Switch to full screen mode
3. Return to window mode

**Expected behavior**
Video output should be same as before going full screen.

**Screenshots**
Before going full screen:
image
After returning from full screen:
image

 - OS: Windows 10
 - 86Box version: v3.0 build 2810
 - Build type: regular

**Additional context**

ts-korhonen:
Summary
=======
## Problem analysis ##
Returning from full screen has not worked properly after changing `WM_MOVE` and `WM_SIZE` to `WM_WINDOWPOSCHANGED`.
There is a check for `SWP_NOSIZE` flag, but returning from full screen doesn't disable it and thus the rendering window remains the size of full screen window and offset. Mouse capture area is likewise misaligned.
## Fix ##
Adding an additional check for variable `user_resize` alongside the existing check for `SWP_NOSIZE` allows the rendering window resize be triggered when exiting full screen.

Checklist
=========
* [x] Closes #1355 

References
==========
The originating window resize event changes: 220fffc4a211c688f3aeaf77038a447a8eb5e295

### DIFF
--- a/src/win/win_ui.c
+++ b/src/win/win_ui.c
@@ -835,7 +835,7 @@ MainWindowProcedure(HWND hwnd, UINT message, WPARAM wParam, LPARAM lParam)
 			doresize = 1;
 		break;
 
-    case WM_WINDOWPOSCHANGED:
+	case WM_WINDOWPOSCHANGED:
 		pos = (WINDOWPOS*)lParam;
 		GetClientRect(hwndMain, &rect);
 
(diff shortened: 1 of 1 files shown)


dhrdlicka:
Thanks!

PR #1368, opened 2021-04-11
Few minor changes on the Intel i450KX

tiseno100:


src/chipset/intel_i450kx.c
```diff
@@ -354,8 +346,8 @@ mc_write(int func, int addr, uint8_t val, void *priv)
     case 0x6d:
     case 0x6e:
     case 0x6f:
-        dev->mc_pci_conf[addr] = ((addr & 0x0f) % 2) ? val : (val & 7);
-        spd_write_drbs(dev->mc_pci_conf, 0x60, 0x6f, 1);
+        dev->mc_pci_conf[addr] = ((addr & 0x0f) % 2) ? 0 : (val & 0x7f);
```
qeeg, line 349: `((addr & 0x0f) % 2)` should probably be changed to `!(addr & 0x01)` for better clarity :P

src/chipset/intel_i450kx.c
```diff
@@ -324,7 +317,6 @@ mc_write(int func, int addr, uint8_t val, void *priv)
 
     case 0x58:
         dev->mc_pci_conf[addr] = val & 2;
-        mem_set_mem_state_both(0xa0000, 0x20000, (val & 2) ? ENABLE_SEGMENT : DISABLE_SEGMENT);
```
qeeg, line 327: Why was this removed?

qeeg:
Looks mostly good, just a couple of nitpicks

src/chipset/intel_i450kx.c
tiseno100: Sure (addr & 1) would be nice just modulo 2 came to my mind as first choice

src/chipset/intel_i450kx.c
tiseno100: On the MC, that register just enables the A0000-BFFFF segment. Just so I can avoid conflicts with the PB. i450KX works in a way 86Box can't properly support.

PR #1382, opened 2021-04-16
OpenGL renderer a.k.a. shader support

ts-korhonen:
Summary
=======
I've been working on shader support via renderer that creates OpenGL context window allowing use of OpenGL library. It is now in a state where it seems to be working while there's still more to be done. Currently it replaces the existing SDL OpenGL option in menus and has no additional UI. I have selected the libretro shader files as a target for compatibility.

Implemented features in current state:
- Rendering in OpenGL (3.3 core)
- Basic shader support for libretro format shader files
  - Single-pass, no previous frames or scaling
  - path is hardcoded (shaders/shader.glsl)

Planned features:
- Menu options & saving them in config
- Reading .glslp files for settings
- Multiple shader passes
- Keeping previous frames

I didn't want to start modifying the UI yet and I opened this draft PR hoping to get some feedback from core developers.
Should this current state be merged into main repo, maybe in the experimental build?
Should it be it's own option and not replace the current SDL OpenGL option like it's doing now?

Checklist
=========
* [ ] Closes #xxx
* [ ] I have discussed this with core contributors already
* [ ] This pull request requires changes to the ROM set
  * [ ] I have opened a roms pull request - https://github.com/86Box/roms/pull/changeme/

References
==========
_Provide links to datasheets or other documentation that helped you implement this pull request._

PR #1386, opened 2021-04-17
Make DEV_BRANCH available in .rc files when building with make

ts-korhonen:
Summary
=======
Defining DEV_BRANCH affects only source files when building with make. In CMake it affects both source files and resource files.

#1382 didn't build as expected while this was missing; the experimental menu option is missing.


OBattler:
Strange that this merged without problems, since I thought I committed the same change myself.

Issue #1160, opened 2020-12-07
Strange bugs in Olivetti machines


Fenix770:
**Describe the bug**
I found three bugs in the three different Olivetti machines.

1.The M19 8088 model does not boot.

2. The M24 machine works, but I cannot use the standard serial mouse even if I specify it in the input options. I tested it with Windows 1.03. Works with the M240.

3. The M240 machine works right, but detect the hard disk drive C as E. It occurs on Olivetti M240 with Olivetti MS-DOS 3.20, but it detects correctly as C drive on the M24.
I tried to add Lastdrive=C  line inside CONFIG.SYS without avail.

Geoworks Detects the B,C, and D drives as 3.5 floppies even but they do not exist, as the only drives configured are a 5.25 360 KB (as A drive) and a 20mb hard disk (in this case E drive, but it should be C).

**To Reproduce**

1.Mount the 20MB hard disk C/H/S 615/4/17 with IBM PC (Xebec) or DTC-5150X controller
2. Execute the Win command in C:\Windows directory

On M24:
3. Windows 1.03 is configured with the Olivetti CGA 640x400 2 color mode and a standard serial mouse but, it is not detected on this machine.

On M240:
3.Windows 1.03 works but, detects drive B,C and D and they should not exist.
4. Start GeoWorks. From C:\geoworks directory, type PCGEOS,
5. Select professional mode
6. Drive letters B,C and D appears as 3.5 floppy drives.

**Screenshots**
![BUG 001](https://user-images.githubusercontent.com/31600563/101413753-9553b000-38c3-11eb-93aa-80525ad23488.png)

The M240 boot with E drive
![BUG 002](https://user-images.githubusercontent.com/31600563/101413793-a7cde980-38c3-11eb-9fc0-494b00d8b3d3.png)

The M240 running GwoWorks
![BUG 003](https://user-images.githubusercontent.com/31600563/101413842-c338f480-38c3-11eb-98bf-7cf2beaebd6b.png)

Here is the Hard Disk image:
[Olivetti 615-4-17.zip](https://github.com/86Box/86Box/files/5655647/Olivetti.615-4-17.zip)

**Desktop (please complete the following information):**
 - OS: Windows 10 ver 2004
 - 86Box version:  7e96d5e8 
 - Build type: Regular

**Additional context**
The M24 mouse setting (Serial mouse) works right in PCem. The other machines were not implemented.

Fenix770:
[...]

PR #1460, opened 2021-05-30
Philips and Olivetti bugfixes

EngiNerd89:
Summary
=======
Fixed Philips P3120 FDC bug: internal FDC controller now works
Fixed Olivetti M300 NVRAM bug: when initialized for the first time, NVRAM prevented internal FDC and HDC controllers from working once CMOS settings were saved.

Checklist
=========
* [X] Closes https://github.com/86Box/86Box/issues/1160

References
==========
N/A

### DIFF
--- a/src/machine/m_xt_philips.c
+++ b/src/machine/m_xt_philips.c
@@ -150,9 +150,6 @@ machine_xt_philips_common_init(const machine_t *model)
 
     pit_ctr_set_out_func(&pit->counters[1], pit_refresh_timer_xt);
 
-    /* On-board FDC cannot be disabled */
-    device_add(&fdc_xt_device);
-    
     nmi_init();
 
     if (joystick_type)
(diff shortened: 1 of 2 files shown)


OBattler (commented):
While this is needed for the Olivettis, it would be best to add its own NVR rather than ch anging the default to 0xff for *all* machines.

EngiNerd89:
I see, I'll then close this pull request and create a new one with an NVRAM specific exception for Olivetti machines (which I didn't at first since before the recent modifications in nvr_at.c the default value *was* 0x00 for all machines), as long as there is still interest in these machines.

Concerning the issue you pointed out (multiple restarts at boot), I never experienced it on my local setup with M300 machines. Conversely, the PCS-46C that was recently added continuously restarts complaining of a Timer 1 error, so yes, 486-class Olivetti machines (as well as other machines of other brands I tried to add, ranging from 8088 to 486) *have* this issue.

PR #1464, opened 2021-06-02
Philips and Olivetti bugfixes (revised)

EngiNerd89:
Summary
=======
Fixed Philips P3120 FDC bug: internal FDC controller now works
Fixed Olivetti M300 NVRAM bug: when initialized for the first time, NVRAM prevented internal FDC and HDC controllers from working once CMOS settings were saved. A specific NVRAM device for Olivetti machines has been created in nvr_at.c

Checklist
=========
* [X] Closes https://github.com/86Box/86Box/issues/1160

References
==========
N/A

### DIFF
--- a/src/include/86box/nvr.h
+++ b/src/include/86box/nvr.h
@@ -92,6 +92,7 @@ extern const device_t at_nvr_device;
 extern const device_t ps_nvr_device;
 extern const device_t amstrad_nvr_device;
 extern const device_t ibmat_nvr_device;
+extern const device_t olivetti_nvr_device;
 extern const device_t piix4_nvr_device;
 extern const device_t ls486e_nvr_device;
 extern const device_t ami_apollo_nvr_device;
--- a/src/machine/m_at_286_386sx.c
+++ b/src/machine/m_at_286_386sx.c
@@ -833,7 +833,7 @@ machine_at_m30008_init(const machine_t *model)
 	return ret;
 
     machine_at_common_init_ex(model, 2);
-    device_add(&ps_nvr_device);
+    device_add(&olivetti_nvr_device);
 
     device_add(&opti283_device);
     device_add(&keyboard_ps2_olivetti_device);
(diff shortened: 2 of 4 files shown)


OBattler:
The Olivetti M300's have the same PIT problem as the 486 (though I need to try with the reverted ROM read timing, maybe that will fix it?), as such they are likely to be removed (but might be kept if someone figures out a way to fix that stuff).

OBattler:
OK, that did not fix them, so the Olivetti M300's and 486 are getting removed.

EngiNerd89:
That's weird, as M300 machines work fine on my end. Maybe these PIT issues are host CPU sensitive? FYI my host machine is a ThinkPad X230 with an i5-3320m

OBattler:
They appear to work fine, but if you notice, they restart themselves a few times on POST, that's because of the PIT problems. The M300's just don't explicitly say what's wrong, unlike the 486.

EngiNerd89:
You're right. When those machines are cold-started, the 86box window resizes 2 times before the POST messages (Resident Diagnostics) is shown. I hadn't payed much attention to this, as I was thinking it was the normal behavior of the VGA adapter, and not a PIT issue. While testing, I also found out the cause of non-working on-board graphics adapters some users were lamenting: when machine names were renamed and directory paths were changed, the path of the on-board VGA BIOS was not changed.

That said, what should we do now? Should I close this commit and create a new one only with fixes for Philips machines, leaving the Olivetti ones in my local repository? Or should this commit get approved first and later, once the decision is finalized, the M300 machines get removed?

OBattler:
> That said, what should we do now? Should I close this commit and create a new one only with fixes for Philips machines, leaving the Olivetti ones in my local repository? Or should this commit get approved first and later, once the decision is finalized, the M300 machines get removed?

You should do the former (but don't actually remove the machines yet as they are removed on my end and you making a removal PR would cause a conflict with my local changes) as the decision is finalized and the machines have already been removed on my end.

PR #1502, opened 2021-06-20
Apparently the typo that was fixed caused new small border glitches, …

TC1995:
…so, reverting it.

Summary
=======

Checklist
=========
* [ ] Closes #xxx

References
==========


qeeg:
Oh come on, surely this is wrong from a hardware behavior point of view...

OBattler:
> 
> 
> Oh come on, surely this is wrong from a hardware behavior point of view...

Yes, I did just find one book about old graphics cards that documents a whole set of extended Trident graphics engine registers that we don't even implement at all, and that are most likely used here, since they include registers needed for Bresenham lines.

PR #1663, opened 2021-09-05
Fix CPU prefetch so that it's disabled on 486 and higher when cache is on, just like the comment near it says

qeeg:
Summary
=======
This change simply brings actual behavior regarding CPU prefetch in line with intended behavior.

Checklist
=========


OBattler:
Closing, because not only it worsens performance on non-Ryzen CPU's, but it outright breaks stuff.

PR #1755, opened 2021-10-17
Make it build in arch linux

andi8086:
Signed-off-by: Andreas J. Reichel <homebase_ar@web.de>

Summary
=======
master does build in arch linux with these changes

Checklist
=========
* libui.a needs SDL linked into it, maybe because of newer gcc 11 on Arch
* SDL2 links now on both Debian And Arch Linux that use different targets for linkage!


andi8086:
Got it working now for both Debian and Arch-Linux with either FindSDL2.cmake
or sdl2-config (providing SDL2::SDL2 or ${SDL2_LIBRARIES} but never both) :)

andi8086:
Sorry, I forgot to take out the two files. Now it shouldnt.

PR #1790, opened 2021-11-08
Initial commit for multilingual UI, and Hungarian translation

laciba96:
Summary
=======
This PR adds a kinda effortless solution to translate the current UI, via separating the RC file into multiple parts. 

2021-11-08_19h17_07

The UI translation after this PR can be added to `src\win\languages` by translating `en-US.rc` and then adding a single line to `86Box.rc` to include the new translation. The language-dependent loading of resources are done with the existing code, and Windows itself, no further changes needed.

The PR also adds a Hungarian translation for 86Box, as first translated language.

Edit: Even if an UI rewrite is in progress, I think the translations for multiple languages probably can re-used after the rewrite.

Checklist
=========
* [ ] Closes #xxx

References
==========
https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-LCID/%5bMS-LCID%5d.pdf


laciba96:
#### A brief introduction to this language system

The basic concept of this language system, that 86Box has a Win32 UI currently, which uses resources to display dialogs, menus, controls, etc.

**What's under the hood?**

The Windows itself allows to define multiple resources on the same name with different languages. This PR uses this feature to provide multilanguage UI with minimal changes.

In order to achieve this the `win/86box.rc` file was split to a localizable, and a non-localizable part. After then the localizable part was translated to Hungarian language.

**Changing the language**

Changing language is very complicated, because needs a heavy planning about how all the possibly localized in-memory contents should change when the language-switch occurs. 

Because of this I've decided about allowing the language-switch only at startup. This means that the selected display language is fixed for the whole run-time.

Originally this system used the Windows's display language, without allowing to differ from it. By that in order to display the Hungarian UI, you had to use Hungarian Windows (or MUI).

**Introducing --lang or -G**

With the recent commits this behavior was changed, so you can choose between any translated languages you wish for 86Box, with the ˙-G <langcode>˙ or ˙--lang <langcode>˙ parameter.

You can find a list about language codes [here](https://www.science.co.il/language/Locale-codes.php).

**Example**

In order to test 86Box in Hungarian, without having Hungarian Windows (or MUI), you can use the following command line.

`86box --lang hu-HU`

**How can I translate 86Box?**

1, Look up for the desired language in the linked table above.
2, Create a copy from `win/languages/en-US.rc` to `win/languages/<your_language-code>.rc`
3, Change the header of the fresh RC file according to the destination language. You can find the constants' name [here](https://www.instantiations.com/vast-support/documentation/92/pr/wwhelp/wwhimpl/common/html/wwhelp.htm#href=stpr608.html&single=true).
 (e.g. for Hungarian I had to use `LANGUAGE_HUNGARIAN,SUBLANG_DEFAULT`, or 0xE,0x1)
4, Also add your name as Translator to the header.
4, Translate all the strings, then save your changes. 
5, Open `win/86box.rc`, and add a new `#include` line, where you link to your fresly created translation.
6, Build 86Box with the changes, and start it with your new translation. Check that every text fits in their place.
7, Done!

PR #1816, opened 2021-11-16
ISA-386C Supports 64MB RAM

Dizzy611:
Summary
=======

The ASUS ISA-386C in addition to supporting 32MB onboard, supports an additional 32MB through a proprietary RAM card using an extended ISA connector marked "S1" on the board. This PR simply edits the machine table to add this extra 32MB to the RAM limit.

I have no idea on implementation details of this, but the manual seems to imply it just ends up mapped directly into RAM banks 2 and 3. So I'm not sure if raising the RAM limit here is the correct fix, but... yeah. Up to y'all. I won't be offended if you reject this PR out of hand :) This would be our highest-RAM 386 system, if added. 

Checklist
=========
* [ ] Closes #xxx

References
==========

According to documentation obtained here:
https://www.win3x.org/uh19/public/motherboard/manual/30218.pdf


Dizzy611:
Updated link for the data about the RAM card
http://www.uncreativelabs.de/th99/m/A-B/30218.htm

PR #1817, opened 2021-11-16
Changeable external icon sets

laciba96:
Summary
=======
This PR extends 86Box's capabilities by adding dynamically changeable external iconsets.

### How to use it:
  Open > Tools > Program Settings... > Iconset:
  
### How it saves it:
  To 86Box.cfg [General] iconset =

### How to add iconset:
  1, Convert your icons to .ICO format. You should have 16x16, 20x20, 24x24, and 32x32 sizes at least for DPI-scaling.
  2, Name your icons as in src\win\icons. Be sure that you have a replacement for every single icon.
  3, Create a new txt file named `iconset.txt`, and write a single line into it - the name you want to see in the combobox.
  4, Choose a short name for your iconset and copy the files to `roms\icons\<your chosen name>\`.
  5, The new iconset should show up in the combobox.
  
### Why `roms\icons\` folder?
Because the romset has to be downloaded for every 86Box instances, so this is the easiest way to share extra icons drawn by the community, without compiling them into the program. Also this allows the check and approve any new icon sets, and the path can be redirected with all the existing infrastructure (e.g. command line parameters) for the rom directory. The rom path currently is well definied for Unix too, but introducing a new path for this could lead to complications in this case.

### Do we have examples?
For example I'll open two PR's on the romset side to show how these iconsets could be created.

The Legacy iconset is identical to the current one, which hopefully will became really legacy when v3.0 is complete.
The Inspired by WinBox iconset is the one, used in the ImGui branch currently.

Since the extra sets does not have hard infrastructure, hopefully there will be more in the future.

### Under the hood:
- The dialog is reused, and added the new strings by reusing the existing ones.
- The (Default) string is constructed from "(" + IDS_2090 + ")"
- The iconsets are explored from the folder returned by `plat_get_icons_path()`, where a FindFirstFile-FindNextFile-FindClose loop will automatically recognizes all the directories
- If the txt file exists, it will be opened as text file, and one line will be read from it as text for the combobox.
- The icon combo has pointers in the CB_ITEMDATA field to strings for the folder name. The display name is constructed from the txt or as fallback from the directory name. The strings are destroyed in WM_DESTROY of the dialog.
- The (Default) entry is stored as icon_set="" which means, that icons should be loaded from resources, while for the other ones the folder name will be stored as icon_set.
- The existing icon table was used for the StatusBar, but by adding all the missing icons, so from now the Settings dialog will also use these. plat_clear_icon_set() will destroy the table, and plat_load_icon_set() will fill, according to icon_set.
- The name of the expected icon files are stored in the icon_files[] table. It can be expanded later when needed.

### Notes:
- You can change the icons dynamically in the folder for developing purposes, but you have to reload the set from the dialog.
- It's not necessary to add such a txt file, but without this the folder's name will be displayed.
- Icons can be missing from your set, but not recommended - in this case the compiled ones will be used instead. 
- The PR can handle if you remove your iconset, but the config still contains it. It'll simply revert to the default one in such cases.
- The iconinfo.txt can hold more than one lines, for the GUI only the first one matters. For example you can add copyright, license, release date, or you can mention the creators' names, etc. after the first line.

Checklist
=========
* [ ] Closes #xxx
  * [X] I have opened a roms pull request 
  https://github.com/86Box/roms/pull/144
  https://github.com/86Box/roms/pull/143

References
==========
N/A


dhrdlicka:
I am not all that convinced about the actual usefulness of this feature. To be fair it only adds to feature creep and doesn't provide any non-negligible benefits.

src/86box.c
```diff
@@ -715,6 +715,9 @@ pc_init(int argc, char *argv[])
 	if (lang_init)
 		set_language(lang_init);
 	
+	/* Load the desired iconset */
+	plat_load_icon_set();
+
```
dhrdlicka, line 720: Should be done in `ui_init` really

src/include/86box/plat.h
```diff
@@ -125,6 +125,11 @@ extern void	plat_vid_reload_options(void);
 extern uint32_t plat_language_code(char* langcode);
 extern void plat_language_code_r(uint32_t lcid, char* outbuf, int len);
 
+extern void plat_clear_icon_set();
+extern void plat_system_icon_set();
+extern void plat_load_icon_set();
+extern void plat_get_icons_path(char* path_root);
+
```
dhrdlicka, line 132: Why is this in the platform module? From what I see, icon sets are not used outside of the UI code, so this seems rather redundant to me.

src/win/languages/en-US.rc
```diff
@@ -284,6 +284,7 @@ END
 #define STR_GLOBAL		"Save these settings as &global defaults"
 #define STR_DEFAULT		"&Default"
 #define STR_LANGUAGE		"Language:"
+#define STR_ICONSET "Iconset:"
```
dhrdlicka, line 287: Indentation; also the dialog name should be changed if it's not solely display language selection anymore.

I wouldn't translate to other languages yet either and wait for the translations to be sure we don't ship some machine translated weirdness in the build. (This reminds me there is no fallback for missing strings in translations.)

src/win/win_icon.c
```diff
@@ -0,0 +1,154 @@
...
+		mbstowcs(wtemp, temp, strlen(temp) + 1);
+		
+		HICON ictemp;
+		ictemp = LoadImageW(NULL, (LPWSTR)wtemp, IMAGE_ICON, x, y, LR_LOADFROMFILE | LR_DEFAULTCOLOR);  
+		if (ictemp)
+		{
+			HICON* helper = &hIcon[icon_files[i].id]; 
+			if (*helper)
+				DestroyIcon(*helper);
+			*helper = ictemp;
```
dhrdlicka, line 147:
```c
// Destroy the existing icon if one
if (hIcon[icon_files[i].id])
    DestroyIcon(hIcon[icon_files[i].id]);
hIcon[icon_files[i].id] = ictemp;
```

This seems cleaner to me and it saves you a variable.

src/win/win_lang.c
```diff
@@ -76,6 +78,81 @@ progsett_fill_languages(HWND hdlg)
...
+	
+	if (combo_index == -1)
+	{
+		combo_index = 0;
+		strcpy(temp_icon_set, "");
+	}
+	
+	SendMessage(icon_combo, CB_SETCURSEL, combo_index, 0);
+}
+
```
dhrdlicka, line 155: Seems a bit strange to put this in the Language dialog; if your intent is to turn it into a global settings dialog then perhaps the file should be renamed.

src/win/win_ui.c
```diff
@@ -78,7 +77,7 @@ extern WCHAR	wopenfilestring[512];
 static wchar_t	wTitle[512];
 static int	manager_wm = 0;
 static int	save_window_pos = 0, pause_state = 0;
-static int	dpi = 96;
+int	dpi = 96;
```
dhrdlicka, line 80: Indentation

src/win/win_ui.c
```diff
@@ -1074,6 +1064,7 @@ MainWindowProcedure(HWND hwnd, UINT message, WPARAM wParam, LPARAM lParam)
 		break;
 
 	case WM_DESTROY:
+	    plat_clear_icon_set();
```
dhrdlicka, line 1067: Indentation

src/unix/unix.c
```diff
@@ -1244,6 +1245,30 @@ plat_language_code_r(uint32_t lcid, char* outbuf, int len)
...
+{
+	return;
+}
+
+void
+plat_get_icons_path(char* path_root)
+{
+	return;
+}
+
```
dhrdlicka, line 1271: No need to stub anything if the icon set code is local to the UI code.

src/win/win_stbar.c
```diff
@@ -878,46 +877,7 @@ StatusBarPopupMenu(HWND hwnd, POINT pt, int id)
...
-    for (i = 176; i < 178; i++)
-	hIcon[i] = LoadImage(hInst, MAKEINTRESOURCE(i), IMAGE_ICON, x, x, LR_DEFAULTCOLOR);
-    for (i = 184; i < 186; i++)
-	hIcon[i] = LoadImage(hInst, MAKEINTRESOURCE(i), IMAGE_ICON, x, x, LR_DEFAULTCOLOR);
-    for (i = 192; i < 194; i++)
-	hIcon[i] = LoadImage(hInst, MAKEINTRESOURCE(i), IMAGE_ICON, x, x, LR_DEFAULTCOLOR);
-    hIcon[232] = LoadImage(hInst, MAKEINTRESOURCE(232), IMAGE_ICON, x, x, LR_DEFAULTCOLOR);
-    for (i = 243; i < 244; i++)
-	hIcon[i] = LoadImage(hInst, MAKEINTRESOURCE(i), IMAGE_ICON, x, x, LR_DEFAULTCOLOR);
+	plat_load_icon_set(hInst);
```
dhrdlicka, line 880: Your `plat_load_icon_set` function does not accept any arguments.

dhrdlicka (changes requested):

src/win/languages/dialogs.rc
```diff
@@ -1,14 +1,17 @@
...
+    PUSHBUTTON      STR_CANCEL, IDCANCEL, 179, 97, 50, 14  
     COMBOBOX        IDC_COMBO_LANG, 13, 18, 213, 22,  CBS_DROPDOWNLIST | CBS_HASSTRINGS
-    AUTOCHECKBOX    STR_GLOBAL, IDC_CHECKBOX_GLOBAL, 13, 50, 217, 8 , WS_DISABLED 
     PUSHBUTTON      STR_DEFAULT, IDC_BUTTON_DEFAULT, 162, 32, 60, 14 
-    LTEXT	    STR_LANGUAGE, 0, 13, 8, 34, 8
+    COMBOBOX        IDC_COMBO_ICON, 13, 50, 213, 22,  CBS_DROPDOWNLIST | CBS_HASSTRINGS
+    PUSHBUTTON      STR_DEFAULT, IDC_BUTTON_DEFICON, 162, 64, 60, 14 
+    AUTOCHECKBOX    STR_GLOBAL, IDC_CHECKBOX_GLOBAL, 13, 82, 217, 8 , WS_DISABLED 
+    LTEXT	    STR_LANGUAGE, 0, 13, 8, 100, 8
+    LTEXT	    STR_ICONSET, 0, 13, 40, 100, 8
```
dhrdlicka, line 14: This breaks the Visual Studio build.
```
D:\a\86Box\86Box\src\win\languages/dialogs.rc(14): warning RC2182: duplicate dialog control ID 0 [D:\a\86Box\86Box\build\src\win\ui.vcxproj]
```

dhrdlicka (changes requested):

src/win/win_stbar.c
laciba96: I've applied the requested changes in the recent commits, and also brought the branch up-to-date.

src/win/win_ui.c
```diff
@@ -1404,6 +1395,9 @@ ui_init(int nCmdShow)
     /* Reset all menus to their defaults. */
     ResetAllMenus();
     media_menu_init();
+	
+	/* Load the desired iconset */
+	win_load_icon_set();
```
dhrdlicka, line 1400: Indentation

src/win/win_progsett.c
```diff
@@ -107,6 +184,10 @@ progsett_settings_save(void)
     /* Language */
     set_language(temp_language);
 
+	/* Iconset */
+	strcpy(icon_set, temp_icon_set);
+	win_load_icon_set(hinstance);
+
```
dhrdlicka, line 190: Indentation + wrong parameters for `win_load_icon_set`

dhrdlicka (changes requested):
Marked some slight oversights, otherwise the PR seems to be fine code-wise. I still don't agree with the feature as whole though

src/win/win_progsett.c
laciba96: Thank you for the checks and comments, I've applied a fix for these parts of the code too, according to indentation.

If I can mention one thing, where this feature will have a really practical use in the near future, is developing the new icon set (in the same style of the main icon). At the Discord server I've demonstrated that it can provide a very rapid developing environment for this task, where it's not necessary to recompile 86Box to see how the new changes will look in the program. It allows to test quickly the icons, and reload them even after the slightest changes, and only compile them into the program itself, when all of them are absolutely ready.

dhrdlicka (commented):
I don't think I have any new feedback anymore

ts-korhonen:
I have to agree with @dhrdlicka being critical about the usefulness of this feature to regular users and the worry about feature creep. I understand it could make testing icons easier, but I don't think it's justified only to avoid having to recompile.

Issue #256, opened 2018-02-04
Add serial passthrough support on host machines with a serial port


MoochMcGee:
This can be used to, say, debug a Windows kernel running in 86Box on one machine, from another machine.

An example of how to program this on Windows is given here: https://www.codeguru.com/cpp/i-n/network/serialcommunications/article.php/c5425/Serial-Communication-in-Windows.htm

andi8086:
What, this is still open??? I need that feature now XD LOL, this is really basic stuff, I might have a look now into it at least on Unix.

andi8086:
Okay, I need a little help to understand how to add hardware.  I would suggest to add a new device called "serial_passthrough" or something like that and on Linux systems, use a virtual console via /dev/ptmx. However I dont understand how to emulator enumerates available hardware... i.e. what I can set in a config file. Where is the mapping between config file keywords and the devices?

ssokolow:
I'm not familiar with the 86box side of things, but I can advise on design decisions for the Linux side of things as well as what other emulators have done.

First, I'd want some examples of the kind of use-cases you're thinking of.

For example, for me, I'd want something equivalent to DOSbox's support for exposing an emulated serial port as a TCP listening socket, so I can run a testing harness for builds of my DOS creations that have been instrumented to treat COM1 as stderr and pause at various points to wait on COM1 so the harness can reliably take screenshots to compare to known good versions for rendered output tests.

ts-korhonen:
Nice to see people interested in contributing! Please feel welcome to join the [Discord server](https://discord.gg/QXK9XTv) to chat with the other developers :)

andi8086:
:) I am developing a DOS Device driver. And this device driver does things via the serial port. On the other side is a server, which will be a linux program. I use 86Box for testing of the DOS driver. But for this, I must feed through the serial data. My idea is to create a new device, named something like "serial_passthrough". And add a line for PC and up like "serial_passthrough = 1,2" in config, which would then attach the pass_through device to COM1 and COM2 if needed. Then in Linux, I would open /dev/ptmx to get a virtual device and pass the data to and from a new virtual console on read and write. Of course DTS/RTS and speed are not handled, the virtual console more behaves like a FIFO, but this way, also debugging of Kernels should be possible! The virtual terminal can be set to a fixed baud rate, whereas the simulated serial device can keep its baud rate.

[...]

PR #1850, opened 2021-11-19
WIP: are/serial - serial Passthrough

andi8086:
Summary
=======
This is WORK IN PROGRESS - Just that people can see what I am doing.

THE COMMITS ARE NOT CLEANED UP YET :dagger: :D

Checklist
=========
* [X] Targets #256 

Explanation
=========
* Add a new serial device type " serial_passthrough "
* If config adds
[Ports (COM & LPT)]
serial1_passthrough = vcon

where "vcon" is the passthrough mode, (there will be different modes for TCP, virtual console, tty, serial, etc...)

the passthrough virtual device is attatched to the specified serial port.
This has been tested on the zenith xt (the one with com port) and the am495 machine.

The passthrough device gets informed by the write_callback of serial_device_t, and then calls a platform specific
function plat_serpt_write.

This way, it is modular and further platform specific implementation can decide how to exactly send and receive the bytes
from and to the host.

### DIFF
--- a/src/86box.c
+++ b/src/86box.c
@@ -63,6 +63,7 @@
 #include <86box/isartc.h>
 #include <86box/lpt.h>
 #include <86box/serial.h>
+#include <86box/serial_passthrough.h>
 #include <86box/keyboard.h>
 #include <86box/mouse.h>
 #include <86box/gameport.h>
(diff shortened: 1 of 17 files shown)


andi8086:
Feed from emulated machine to host works via the vcon option (/dev/ptmx => /dev/pts/XY)

andi8086:
Reverse direction in vcon mode is working. Status register of serial port is updated according to the file descriptor status.

andi8086:
Implemented Loopback mode, so that pass through is not called if MCR & 0x10

PR #1855, opened 2021-11-20
Implementation of global-defaults infrastructure

laciba96:
Summary
=======
This PR completes the previous two PRs (language and iconset), and Program Settings dialog by implementing the system of global defaults.

### About the global defaults system
In the current 86Box config system everything is stored locally in 86Box.cfg (per-VM). This PR don't want to radically change it, but offers a solution to those who have multiple VMs.

If you want to choose your favorite language eg. pt-BR for every VMs, in the old-way you have to set-up it for every VM.
After this PR you can define it as default for new VMs, so when the language= line is missing from the config, or the VM does not have config, these user-specified default will be loaded (eg. pt-BR).

The PR does not prevent the user to change the language on per-VM basis, also the global-defaults does not affect VMs, with definied local settings. The universal defaults will be also unchanged, e.g. the Default button still reverts the language to en-US, since this is the universal default.

### Where these settings are stored?

According to previous discussions at Discord, the storages would be there:

 * Windows: HKCU\Software\86Box\Preferences
 * Linux: ~/.86Box
 * macOS: ~/Library/Application Support/86Box
 
For Windows it is implemented with this PR, for Linux and macOS it provides a skeleton for the platform-dependent implementation. Since the last to system currently does not have the new features (language and iconset), these platform-dependent functions can be implemented later.
 
 ### Under the hood
 
 There are low-level storage management (e.g. reading a single key-value pair, etc.) in plat.h, win_gconf.c, and unix_gconf.c. 
 
 There are high-level config loading methods like config_load_gconf(), and config_save_gconf() in config,c, and config.h.
 
 The config_save_gconf() is only called from Program Settings dialog, config_load_gconf() when new config file will be created, and the single reader functions are called when the necessary line is missing from the config.
 
 There are no translation-changing UI modifications with this PR, the existing global-defaults checkbox got allowed, and no more.

Checklist
=========
* [ ] Closes #xxx

References
==========
N/A


dgurney:
Leaving it here to so it's more easily on record: I think on Linux it would be better to adhere to [XDG base directory specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), i.e. use $XDG_CONFIG_HOME/86Box/ (by default ~/.config)

jvernet:
For MacOsX, it should be ~/Library/Preferences and use the MacOsX way to store Preferences

dhrdlicka:
Personally I think we should first agree on how we even want to use a global config. I am worried that the idea of default settings which can be overridden on a VM-basis, while allowing for great customizability, might end up being quite confusing for the user. Are there even any legitimate scenarios where a user would want to override language or icon set settings for a single VM? Not to mention this will inevitably lead to users asking why their one VM is using a different language/iconset/whatever even though they changed it in the GUI.

Personally I think a better approach would be to simply have separate local and global settings without any overlap -- a configuration either applies to one or all VMs. Obviously the current system can't be easily migrated to such global settings, but considering it's minor stuff such as language, icon set or renderer preferences, I don't think it will be such a big problem to just force everybody to reconfigure this for once.

Furthermore I think that this whole branching out to use registry on Windows, files on Unix etc. is completely unnecessary and only creates more points of confusion. It might be better to just use a file on all platforms that's just saved to a different location on each OS, so there is a single global settings implementation used everywhere.

laciba96:
The questions mentioned can be used as a good start for a discussion. 
Since currently this PR is outdated, and offered global settings for only two UI-features, I've converted it to draft.

The two approach has its own pros and cons, and we have 86Box characteristics which cannot be taken apart. Since it has no bet currently, and this two settings are not much relevant, I think we can discuss every detail.

As for the two-levelled approach, totally agree it allows high flexibility, but this high flexiblity also can be confusing as @dhrdlicka mentioned. Also the bunch of local settings, which overrides the global settings, can make a config mess, where the global settings has almost no effect. With this two settings, I think the only real-life example for the per-VM language/iconset, that you can adjust 86Box's language to the Guest OS's language, but that's a rather rare example.

Compared to this, the other solution for the user it's a straightforward - click OK in a single dialog, and every VM changes the relevant settings. However in 86Box, a main characteristic is that every VM is a seperate process. If we want to follow this one-step global settings route, we should implement an interprocess global configuration system, which affects all the VMs - which in itself is a big challenge to implement (race conditions in the storage system, the communication itself, and so on). 

When I've made this PR, I've compared the complexity of these two solutions, and I tried the easier route. I don't know the future plans with such a global storage system, so I also don't know that my implementation was suitable for these plans, or not. While I think for the mentioned two settings it was enough, if this feature is not usable later on, it'll just only creates problems with legacy registry keys, and so on, therefore I don't suggest anything related to this.

Another thing that was discussed earlier, by others to is where to store the data. This again marks more questions about the future plans. As I know registry is multiprocess/thread-safe, so it can be used even for the second solution - but removes the current portability of 86Box (it can run from a USB stick, or network share too). By using files the portablility remains but race conditions can occur in the second implementation (what process locks the file first, and for how much time). Also combined approaches can be considered too.

It seems to me on Linux and MacOS the config file solution is the common, but the place should be discussed. Under Windows there are multiple approaches which can be discussed again by pros and cons. 

Another opinion, that maybe we can keep the checkbox in Preferences locked in the Win32 UI, since the UI rewrite is under rapid development, and maybe Qt has some kind of internal solution to create a proper global storage system without too much effort, which solves all the problems mentioned above.

PR #1860, opened 2021-11-21
WIP: Serial Passthrough: Initial implementation of device

andi8086:
Started from scratch again and cloned from mouse_serial to better conform to serial logic.

Targets #256


jriwanek:
Work continues on a new branch

PR #2008, opened 2022-01-24
Minor update ru-RU.rc

KeishiX:
Summary
=======
Fixed a couple strings.

Checklist
=========
* [ ] Closes #xxx

References
==========


src/win/languages/ru-RU.rc
```diff
@@ -295,7 +295,7 @@ END
 #define STR_MACHINE		"Системная плата:"
 #define STR_CONFIGURE		"Настройка"
 #define STR_CPU_TYPE		"Тип ЦП:"
-#define STR_SPEED		"Частота:"
+#define STR_SPEED		"Speed:"
```
lemondrops, line 298: The space for the CPU speed string will be expanded in the future, so there is no need to change the string back.

lemondrops (changes requested):

src/win/languages/ru-RU.rc
KeishiX: That's ok, but there's another issue. This string is also used for CD-ROM speed option for which it is incorrect.

lemondrops: Then, I guess it should be translated as "Скорость:" for now. While it will be clipped on the machine page, it is going to be fixed soon, anyway. The part about the string being used in two different context is noted and passed over to the rest of the development team.

KeishiX: OK, great. I changed it as you suggested. I hope there's no other objections about PR.

PR #2026, opened 2022-01-30
Openal changes

jriwanek:
Summary
=======
Allows openAL support to be disabled
A couple of consistency improvements to the github actions build

Checklist
=========
* [ ] Closes #xxx

References
==========


.github/workflows/cmake.yml
```diff
@@ -207,7 +210,13 @@ jobs:
...
           -D CMAKE_C_COMPILER=gcc-11 -D CMAKE_CXX_COMPILER=g++-11
       - name: Build
-        run: cmake --build build --target install
+        run: cmake --build build
+      - name: Generate package
+        run: cmake --install build --prefix ./build/artifacts ${{ matrix.build.strip }}
+      - uses: actions/upload-artifact@v2
+        with:
+          name: '86Box${{ matrix.build.slug }}-Linux-x86_64-gha${{ github.run_number }}'
+          path: build/artifacts/**
```
dhrdlicka, line 219: Nitpick: I would avoid creating any Linux artifacts that are not self-contained and statically linked with all dependencies, otherwise the binary is useless on pretty much anything that is not the host distro.

src/CMakeLists.txt
```diff
@@ -87,9 +87,11 @@ if(APPLE)
     target_link_libraries(86Box Freetype::Freetype)
 endif()
 
-find_package(OpenAL REQUIRED)
-include_directories(${OPENAL_INCLUDE_DIR})
-target_link_libraries(86Box ${OPENAL_LIBRARY})
+if(OPENAL)
+    find_package(OpenAL REQUIRED)
```
dhrdlicka, line 91: Perhaps the `find_package` call could be not `REQUIRED` and just have `OPENAL` be forced off if it's not found?

dhrdlicka (commented):
I know this is already merged, but I thought I would leave some notes on this

.github/workflows/cmake.yml
jriwanek: Yeah, I'm just cleaning stuff up, I recall the conversation about static linking and fully agree with you too. But that's for another day, unless you want to beat me to it.

src/CMakeLists.txt
jriwanek: I'm really not all that good with cmake build scripts, so if you wanna, go right ahead, it'd be very welcome. (:

PR #2198, opened 2022-03-06
improve Finnish translation

ziplantil:
This is an improvement for the 86Box Finnish translation (both Win32 and Qt). I tried to synchronize the changes between the two, although I might've missed some.

Something I noticed is that there are more Win32 strings than Qt ones. For example, the Win32 version has separate strings for tape drives, cartridges, floppies etc. but there's only one set of strings for Qt. For the latter, the first matches had been picked, which meant that the Qt translation used the tape drive strings for all media. That sounded odd, because floppy drive "eject" translated to "remove from cassette tray". I tried to change the Qt translation to be fairly media-agnostic.

### DIFF
--- a/src/qt/languages/fi-FI.po
+++ b/src/qt/languages/fi-FI.po
@@ -29,7 +29,7 @@ msgid "&Hide status bar"
 msgstr "&Piilota tilapalkki"
 
 msgid "Hide &toolbar"
-msgstr "Hide &toolbar"
+msgstr "Piilota &työkalupalkki"
 
 msgid "&Resizeable window"
 msgstr "&Salli koon muuttaminen"
@@ -38,7 +38,7 @@ msgid "R&emember size && position"
 msgstr "&Muista koko ja sijainti"
 
 msgid "Re&nderer"
-msgstr "&Renderöijä"
+msgstr "&Hahmonnin"
 
 msgid "&SDL (Software)"
 msgstr "&SDL (ohjelmistopohjainen)"
@@ -59,7 +59,7 @@ msgid "Specify dimensions..."
 msgstr "&Määritä koko..."
 
 msgid "F&orce 4:3 display ratio"
-msgstr "Pakota 4:3 näyttösuhde"
+msgstr "Pakota 4:3-näyttösuhde"
 
 msgid "&Window scale factor"
 msgstr "&Ikkunan kokokerroin"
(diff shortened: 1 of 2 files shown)


src/qt/languages/fi-FI.po
```diff
@@ -143,16 +143,16 @@ msgid "&Average"
...
 msgid "CGA/PCjr/Tandy/E&GA/(S)VGA overscan"
-msgstr "CGA/PCjr/Tandy/E&GA/(S)VGA &yliskannaus"
+msgstr "CGA/PCjr/Tandy/E&GA/(S)VGA-&yliskannaus"
 
 msgid "Change contrast for &monochrome display"
 msgstr "&Muuta harmaavärinäytön kontrastia"
 
 msgid "&Media"
-msgstr "&Media"
+msgstr "&Levyt"
```
ts-korhonen, line 152: Not every media is a disk

ts-korhonen:
> For example, the Win32 version has separate strings for tape drives, cartridges, floppies etc. but there's only one set of strings for Qt.

Qt-port is still relatively new and the translations are automatically generated from Win32, disambiguation is a feature that might be implemented on Qt too.
On that note we still haven't manually edited .po files as you've done. This would make synchronizing them more difficult and need's some discussion among the developers.

src/qt/languages/fi-FI.po
```diff
@@ -38,7 +38,7 @@ msgid "R&emember size && position"
 msgstr "&Muista koko ja sijainti"
 
 msgid "Re&nderer"
-msgstr "&Renderöijä"
+msgstr "&Hahmonnin"
```
ts-korhonen, line 41: While a correct term, this might be fairly unknown to common users.

src/qt/languages/fi-FI.po
```diff
@@ -182,19 +182,19 @@ msgid "&Help"
...
-msgstr "&Tietoja 86Box:sta..."
+msgstr "&Tietoja 86Boxista..."
 
 msgid "&New image..."
-msgstr "&Uusi kasettikuva..."
+msgstr "&Uusi levykuva..."
 
 msgid "&Existing image..."
-msgstr "&Olemassaoleva kasettikuva..."
+msgstr "&Valmis levykuva..."
```
ts-korhonen, line 194: This changes the meaning from "existing" to "ready".

src/qt/languages/fi-FI.po
```diff
@@ -251,7 +251,7 @@ msgid "&75 fps"
 msgstr "&75 ruutua/s"
 
 msgid "&VSync"
-msgstr "&VSync"
+msgstr "P&ystytahdistus"
```
ts-korhonen, line 254: Another term, while correct, might not be well known by common users.

src/qt/languages/fi-FI.po
ziplantil: This is another case where it could go both ways. I personally think it's more important to have the "proper" term, even if it is somewhat unknown.

src/qt/languages/fi-FI.po
ts-korhonen: Yes, the length is the reason for shortening to "Media" as in "Tallennusmedia". Many other translations are also affected by not having enough space in the Win32 UI where the translation was originally done. The UI has had some rework after that though and probably now fits longer words.

ziplantil:
> disambiguation is a feature that might be implemented on Qt too.
I could look into this - I have experience working with gettext, which is what .po files are for. It'd probably have to be a separate PR, though...

> On that note we still haven't manually edited .po files as you've done. This would make synchronizing them more difficult and need's some discussion among the developers.
I'll try to get the .po file to automatically generate from the .rc for the new changes.

ts-korhonen:
> I'll try to get the .po file to automatically generate from the .rc for the new changes.

No need, this is fine. It's probably best to wait a bit because Qt will become the default UI on Windows too, maybe already in the next release. So this .rc / .po split needs to be solved soon anyways.

ts-korhonen (approved):

PR #2214, opened 2022-03-11
qt: Add support for Haiku

Cacodemon345:
Summary
=======
* OpenAL backend is enabled by default on Haiku
* bswap include now makes use of builtins if they exist
* Use a local fgetws method for Haiku to avoid crashes
* OpenGL renderers are completely disabled on Haiku (they do not work)
* Software renderer is a widget, not a window on Haiku (fixes blank screen)
* Add option to disable System MIDI at compile time

Checklist
=========

References
==========
The BeOS/Haiku API documentation.


dhrdlicka:
I am not too convinced whether it's really worth mainlining.

Cacodemon345:
I mean the API has been unchanged from that of BeOS and won't be changed before R1.

PR #2241, opened 2022-03-20
fluidsynth: Fallback to system soundfont if one isn't specified

Cacodemon345:
Summary
=======
fluidsynth: Fallback to system soundfont if one isn't specified.

Checklist
=========

References
==========
None.

### DIFF
--- a/src/sound/midi_fluidsynth.c
+++ b/src/sound/midi_fluidsynth.c
@@ -5,6 +5,9 @@
 #    include <stdlib.h>
 #    include <string.h>
 #    include <wchar.h>
+#    ifdef __unix__
+#        include <unistd.h>
+#    endif
 
 #    include <86box/86box.h>
 #    include <86box/config.h>
@@ -248,7 +251,12 @@ fluidsynth_init(const device_t *info)
 
     data->synth = f_new_fluid_synth(data->settings);
 
-    char *sound_font = (char *) device_get_config_string("sound_font");
+    const char *sound_font = (char *) device_get_config_string("sound_font");
+#    ifdef __unix__
+    if (!sound_font || sound_font[0] == 0)
+        sound_font = (access("/usr/share/sounds/sf2/FluidR3_GM.sf2", F_OK) == 0 ? "/usr/share/sounds/sf2/FluidR3_GM.sf2" :
+                      (access("/usr/share/soundfonts/default.sf2", F_OK) == 0 ? "/usr/share/soundfonts/default.sf2" : ""));
+#    endif
     data->sound_font = f_fluid_synth_sfload(data->synth, sound_font, 1);
 
     if (device_get_config_int("chorus")) {


rderooy:
FYI, fluidsynth docs even specify what should be the default soundfont location. Look for synth.default-soundfont
https://www.fluidsynth.org/api/settings_synth.html

Cacodemon345:
It will cause an error message to be printed on the console instead if the soundfont it refers to can't be found. Hence why it is done the way it is now.

dhrdlicka:
Why not use said setting instead of hardcoding the path though?

Cacodemon345:
Other projects do not seem to use the system soundfont with FluidSynth.

rderooy:
DOSBox-X does it like this also, more-or-less.
from `src/gui/midi_synth.h`
```h
#if defined (WIN32)
                        // default for windows according to fluidsynth docs
                        if (FILE *file = fopen("C:\\soundfonts\\default.sf2", "r")) {
                                fclose(file);
                                sf = "C:\\soundfonts\\default.sf2";
                        } else if (FILE *file = fopen("C:\\DOSBox-X\\FluidR3_GM.sf2", "r")) {
                                fclose(file);
                                sf = "C:\\DOSBox-X\\FluidR3_GM.sf2";
                        } else if (FILE *file = fopen("C:\\DOSBox-X\\GeneralUser_GS.sf2", "r")) {
                                fclose(file);
                                sf = "C:\\DOSBox-X\\GeneralUser_GS.sf2";
                        } else {
                                LOG_MSG("MIDI:synth: Specify .SF2 sound font file with midiconfig=");
                                return false;
                        }
#else
                        // Default on "other" platforms according to fluidsynth docs
                        // This works on RH and Fedora, if a soundfont is installed
                        if (FILE *file = fopen("/usr/share/soundfonts/default.sf2", "r")) {
                                fclose(file);
                                sf = "/usr/share/soundfonts/default.sf2";
                        // Ubuntu and Debian don't have a default.sf2...
                        } else if (FILE *file = fopen("/usr/share/sounds/sf2/FluidR3_GM.sf2", "r")) {
                                fclose(file);
                                sf = "/usr/share/sounds/sf2/FluidR3_GM.sf2";
                        } else if (FILE *file = fopen("/usr/share/sounds/sf2/GeneralUser_GS.sf2", "r")) {
                                fclose(file);
                                sf = "/usr/share/sounds/sf2/GeneralUser_GS.sf2";
                        } else {
                                LOG_MSG("MIDI:synth: Specify .SF2 sound font file with midiconfig=");
                                return false;
                        }
#endif
```
DOSBox-staging has its code to find a soundfont in `src/midi/midi_fluidsynth.cpp` which is a bit more involved.
https://github.com/dosbox-staging/dosbox-staging/blob/main/src/midi/midi_fluidsynth.cpp

PR #2255, opened 2022-04-02
linux pcap fix

goeiecool9999:
Summary
=======
This change fixes a difference between pcap on windows and pcap on linux where pcap_next blocks until a new packet arrives.
It may also improve mac OS although this patch uses `ifdef __linux__` because I cannot test on Mac OS.

Checklist
=========
* [ ] Closes #xxx

References
==========
PCAP linux manual page, specifically the section about "packet buffer timeout"
[This github issue goes into more detail](https://github.com/the-tcpdump-group/libpcap/issues/572)

### DIFF
--- a/src/network/net_pcap.c
+++ b/src/network/net_pcap.c
@@ -122,7 +122,7 @@ static const unsigned char
 			*(*f_pcap_next)(void *,void *);
 static int		(*f_pcap_sendpacket)(void *,const unsigned char *,int);
 static void		(*f_pcap_close)(void *);
-#ifdef __linux__
+#ifndef _WIN32
 static int              (*f_pcap_setnonblock)(void*, int, char*);
 #endif
 static dllimp_t pcap_imports[] = {
@@ -135,7 +135,7 @@ static dllimp_t pcap_imports[] = {
   { "pcap_next",	&f_pcap_next		},
   { "pcap_sendpacket",	&f_pcap_sendpacket	},
   { "pcap_close",	&f_pcap_close		},
-#ifdef __linux__
+#ifndef _WIN32
   { "pcap_setnonblock",	&f_pcap_setnonblock	},
 #endif
   { NULL,		NULL			},
@@ -388,7 +388,7 @@ net_pcap_reset(const netcard_t *card, uint8_t *mac)
 	pcap_log(" Unable to open device: %s!\n", network_host);
 	return(-1);
     }
-#ifdef __linux__
+#ifndef _WIN32
     if (f_pcap_setnonblock((void*)pcap, 1, errbuf) != 0)
         pcap_log("PCAP: failed nonblock %s\n", errbuf);
 #endif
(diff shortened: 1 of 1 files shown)


CraigAB69:
Mate,
I applied your patch and recompiled, ICMP seems to be working, but NetBeui is still not working.
I can ping devices on the network but not the host IP address. Or any device bound to the same host's physical card.  Even though they appear in the arp cache on the guest.

goeiecool9999:
> I can ping devices on the network but not the host IP address. Or any device bound to the same host's physical card. Even though they appear in the arp cache on the guest.

Yeah I think this is a pcap limitation on Linux. In my testing different physical machines were reachable but not the host. Haven't tested other pcap 86box instances but I'm not surprised they don't work. It's because Linux views these packets as outbound on the interface and thus doesn't bother processing them for anything local to the machine. As a workaround for contacting the host I found that setting a route for the host IP to my physical router restored connectivity. If you have more than just the host you can try the entire subnet. Not ideal but better than nothing.

Cacodemon345:
`pcap_setnonblock` is available on macOS. It is safe to remove that ifdef.

goeiecool9999:
> `pcap_setnonblock` is available on macOS. It is safe to remove that ifdef.

I'll make it `#ifndef _WIN32` just to make sure it doesn't mess with windows.

goeiecool9999:
> > I can ping devices on the network but not the host IP address. Or any device bound to the same host's physical card. Even though they appear in the arp cache on the guest.
> 
> It's because Linux views these packets as outbound on the interface and thus doesn't bother processing them for anything local to the machine.

[This npcap issue](https://github.com/nmap/npcap/issues/544) shows more details on this. [This stackoverflow question](https://stackoverflow.com/questions/42608576/pcap-sendpacket-dosent-work-on-npcap-loopback-adapter)  also has useful info

CraigAB69:
Sorry mate, I was referring to NBT, NetBIOS over TCPIP.
As far as I understand it, it should be working.

OBattler:
I actually wonder if the blocking is an issue on Windows as well - our PCap code was written a very inexperience developer back in 2014 and the only change since then was another developer (waltje) moving it to a separate thread in 2017. So it's possible that this fix may apply on Windows as well, and the fact it currently works on Windows is by pure chance.

goeiecool9999:
Yes, I think the structure is pretty flawed. Incoming packets and outgoing packets should both be on their own thread in my opinion. In that case you could let the function block again and leave out the arbitrary 10ms sleep in the thread. That would be the proper way to do it.

OBattler:
I just looked at the old code in the PCem-X repository from 2015... and it turns out we did non-blocking back then. I'm not quite sure when we removed that and why.

Also, the whole network code needs heavy rewriting, what's blocking it right now is the fact TC1995 ported the AMD PCNet series of NIC's improperly and that's standing in the way of the RX/TX restructure.

PR #2299, opened 2022-04-22
Update RPM spec file

rderooy:
Summary
=======
- Add a release date variable, and use that to update the release date in the metainfo file
- Update release to 3.4.1

### DIFF
--- a/src/unix/assets/86Box.spec
+++ b/src/unix/assets/86Box.spec
@@ -12,6 +12,7 @@
 # After a successful build, you can install the RPMs as follows:
 #  sudo dnf install RPMS/$(uname -m)/86Box-3* RPMS/noarch/86Box-roms*
 
+%global date 2022-04-22
 %global romver 20220319
 
 Name:		86Box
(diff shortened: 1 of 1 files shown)


src/unix/assets/86Box.spec
```diff
@@ -119,5 +122,5 @@ popd
 %{_bindir}/roms
 
 %changelog
-* Sat Mar 19 2022 Robert de Rooy <robert.de.rooy[AT]gmail.com> 3.3-1
-- Initial RPM release
+* Fri Apr 22 2022 Robert de Rooy <robert.de.rooy[AT]gmail.com> 3.4.1-1
```
jriwanek, line 125: Would be a nice addition if this used the date global and Version. Don't know if that's possible though.

rderooy: No, unfortunately it is not.

rderooy: what some other projects do, is to run some script before release on the project that updates all the version and date references everywhere.

PR #2396, opened 2022-06-19
qt: Build improvements

Cacodemon345:
Summary
=======
* [x87_ops: Use __asm for consistency with the other inline assembly block](https://github.com/86Box/86Box/commit/ce34c4cb16219d03da2f30b712f745b5cfbf7b9e)
* [qt: Add and use local copies of Vulkan headers in case the system lacks them](https://github.com/86Box/86Box/commit/f17888c4d6f1bf0c1dd914c4717e4ec44461480e)

Checklist
=========

References
==========
None.


dhrdlicka:
Not quite sure what's the purpose of dumping the entire set of Vulkan headers into the 86Box source? I would have thought it's reasonable to assume that a platform that ships Qt with Vulkan support enabled would have also shipped Vulkan headers.

PR #2418, opened 2022-07-01
Add support for using Hercules as a second monochrome video card

Cacodemon345:
Summary
=======
[Add support for using Hercules as a second monochrome video card](https://github.com/86Box/86Box/commit/8be9dc82a315939a0d91b1bfc1fc11b66c306269)

Checklist
=========

References
==========
None.


dhrdlicka (changes requested):
The whole way you added the secondary view seems like a giant hack to me. This might be good enough as a proof of concept, but is not something worth merging into the main line just yet.

PR #2451, opened 2022-07-12
qt: Fix performance regression

Cacodemon345:
Summary
=======
qt: Fix performance regression

Checklist
=========

References
==========
None.

### DIFF
--- a/src/include/86box/video.h
+++ b/src/include/86box/video.h
@@ -225,30 +225,32 @@ extern int  video_card_get_flags(int card);
 extern int 	video_is_mda(void);
 extern int 	video_is_cga(void);
 extern int 	video_is_ega_vga(void);
-extern void	video_inform(int type, const video_timings_t *ptr);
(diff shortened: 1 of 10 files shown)


src/include/86box/video.h
```diff
@@ -225,30 +225,32 @@ extern int  video_card_get_flags(int card);
 extern int 	video_is_mda(void);
 extern int 	video_is_cga(void);
 extern int 	video_is_ega_vga(void);
-extern void	video_inform(int type, const video_timings_t *ptr);
+#define video_inform(type, video_timings_ptr) video_inform_monitor(type, video_timings_ptr, monitor_index_global)
```
dhrdlicka, line 228: Should probably move all the defines together to keep the header tidy.

dhrdlicka (changes requested):

src/include/86box/video.h
Cacodemon345: Done.

PR #2469, opened 2022-07-16
Set default IRQ for NE2000 to 3

Dizzy611:
Increases compatibility with Windows 9x and other operating systems that have a limited selection of IRQs for this device (usually IRQs <= 9).

Summary
=======
This is the smallest and most self-describing PR I think I've ever written. Simply changes the default interrupt for the NE2000 to 3 instead of 10. I'm not sure if 10 was chosen because it was a common default jumper setting on the card or something, but Windows 9x expects 3, 5, 7, or 9 iirc as interrupts for this card, and there are almost certainly other old OSes that are just as picky. A lot of people had the IRQs for their network cards set to 3 anyway, so it would have been a common end-user step to set the jumper to IRQ 3 on the card.

Checklist
=========
* [ ] Closes #xxx

References
==========
None


jriwanek (approved):

rderooy:
Great.... Now there is close to a 100% chance of a resource conflict by default with COM2, which almost every PC back in the day had.

Having done quite a bit of testing with Win 95 and NE2000, I can tell you that the NE2000 driver supplied with Windows 95 does support IRQ10 just fine.

image


Now if you would have said that the card may not work in an XT with only one PIC when set to IRQ10, then yes.

OBattler:
The card is not XT-compatible, anyway, since it's ISA16. For XT, there's the NE1000.

dhrdlicka:
This change also breaks configurations where the IRQ wasn't changed from the default setting.

PR #2514, opened 2022-07-26
Add the FM synth driver select option to the UI

lemondrops:
Summary
=======
Adds an option to select between Nuked OPL and YMFM as the FM synth backend to the Sound page of the settigns dialog.
A groupbox with radio buttons was used to show to the user what the two options are, which is harder to do with a checkbox. Ideas on alternative ways to design it are welcome.

Screenshots
===========
Qt UI:

image

Win32 UI:

image

Checklist
=========


elyosh:
Maybe it would be helpful for users to add a description like Nuked (more accurate) and YMFM (faster) ?

dhrdlicka:
Just a nitpit; I would stack the checkboxes vertically rather than horizontally.

PR #2541, opened 2022-07-30
qt: Global option in preferences to use usr_path as the file image open directory

cold-brewed:
Summary
=======
On macOS, after launching 86Box from finder the image file selection directory defaults to `/`. On Windows and Linux, qt defaults to what is essentially the `$CWD` of the application. There are a few reasons for the macOS difference, mostly just the way application bundles are structured and the way they are launched.

This PR adds an option to try and mimic the Windows / Linux behavior on macOS . When this option is selected any media selection dialog will default to the `usr_path` directory.

The option appears to have no effect on Windows or Linux because open dialogs are in `$CWD` by default. Additionally, current behavior can be maintained by not selecting the option.

Tested on: macOS, linux, windows 10

**Note:** A side effect of this PR is making the preferences dialog wider to display the text for the option (handled automatically by qt). I believe at least one person on discord was requesting something to that effect in order to accommodate longer strings in various translations. 

I think a review by @Cacodemon345 would be good for the qt aspect. Since I had to add another config option and make changes to various files to accommodate I'd like @OBattler to give it a look as well.

### Example

Preferences window

windows prefs

macos prefs

Checklist
=========

References
==========
N/A

### DIFF
--- a/src/86box.c
+++ b/src/86box.c
@@ -183,6 +183,7 @@ int confirm_save = 1;				/* (C) enable save confirmation */
 int	enable_discord = 0;			/* (C) enable Discord integration */
 int pit_mode = -1;				/* (C) force setting PIT mode */
 int fm_driver = 0;				/* (C) select FM sound driver */
+int open_dir_usr_path = 0;                      /* default file open dialog directory of usr_path */
 
 /* Statistics. */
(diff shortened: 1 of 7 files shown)


OBattler:
It seems fine to me.

PR #2557, opened 2022-08-02
sound: Refactor OpenAL backend

Cacodemon345:
Summary
=======
sound: Refactor OpenAL backend

Also adds the ability to create extra sound sources for both OpenAL and XAudio2.

Checklist
=========

References
==========
None.


src/include/86box/sound.h
```diff
@@ -71,6 +77,8 @@ extern void closeal(void);
 extern void inital(void);
 extern void givealbuffer(void *buf);
 extern void givealbuffer_cd(void *buf);
+extern void givealbuffer_source(void *buf, al_source* src, int size, int freq);
+extern al_source* al_create_sound_source(uint32_t freq, uint32_t buffer_size_mono);
```
dhrdlicka, line 81: Given that you are already messing with these parts, would it be possible to change the function names to something a bit more generic?

dhrdlicka (changes requested):

src/include/86box/sound.h
Cacodemon345: al makes more sense since it's the audio library we are interfacing with.

dhrdlicka: not if XAudio2/FAudio is enabled...

Cacodemon345: They count as audio libraries as well.

lemondrops: Except that `al` in these functions' names refers to Open**AL** specifically, not just any audio library.

OBattler: Well, it turns out this PR has more problems: My Windows 98 SE VM drops from 100% to 92% on the POST screen when the NCR 53c8xx SCSI BIOS runs, with this PR, so my intuition that this was going to slightly slowdown the emulator overall, has been correct.

Cacodemon345:
I suspect this will cause performance issues down the line when the extra sources feature start to be used. Closing.

PR #2576, opened 2022-08-05
use zip for roms to prevent clobbering

rderooy:
Summary
=======
With both the main and roms package having the exact same source package filenames, one clobbers the other and the build fails. Prevent this by using the zip package instead for the ROMs.


rderooy:
Since the main package got updated to 3.7.1, but ROM's did not, clearly the ROM's package needs to be separate from the main package version.

PR #2636, opened 2022-08-23
qt: Shortcut fixes

### LINKED ISSUE #2626: The CTRL+F11 and CTRL+F12 key combinations do not work in fullscreen mode
**Describe the bug**
Keyboard shortcuts CTRL+F11 (screenshot) and CTRL+F12 (reload) do not work in fullscreen mode. This bug was introduced in build 3501. In build 3495 everything works fine.

**To Reproduce**
1. Start the machine in 86Box.
2. Switch to fullscreen mode.
3. Press CTRL+F11 (screenshot) or CTRL+F12 (reload).
4. See error.

**Expected behavior**
The CTRL+F11 and CTRL+F12 key combinations should work fine in fullscreen mode.

 - OS: Windows 10 Pro x64 v21H2
 - 86Box v3.7.1 x64 Build 4060
 - Build type: regular, old recompiler.

Cacodemon345:
Summary
=======
qt: Shortcut fixes

* Make Pause key pause/resume the emulator when both Alts are held, passthrough to emulated machine otherwise.
* Make screenshot taking and C+A+D shortcuts work on fullscreen

Checklist
=========
* [X] Closes #2626

References
==========
None.

### DIFF
--- a/src/qt/qt_mainwindow.cpp
+++ b/src/qt/qt_mainwindow.cpp
@@ -1615,7 +1615,18 @@ void MainWindow::keyPressEvent(QKeyEvent* event)
     if (send_keyboard_input && !(kbd_req_capture && !mouse_capture && !video_fullscreen))
     {
         // Windows keys in Qt have one-to-one mapping.
-        if (event->key() == Qt::Key_Super_L || event->key() == Qt::Key_Super_R) {
+        if (event->key() == Qt::Key_Pause && !keyboard_recv(0x38) && !keyboard_recv(0x138)) {
+            if ((keyboard_recv(0x1D) || keyboard_recv(0x11D)) {
+                keyboard_input(1, 0x46);
+            } else {
+                keyboard_input(0, 0xE1);
+                keyboard_input(0, 0x1D);
+                keyboard_input(0, 0x45);
+                keyboard_input(0, 0xE1);
+                keyboard_input(1, 0x1D);
+                keyboard_input(1, 0x45);
+            }
+        } else if (event->key() == Qt::Key_Super_L || event->key() == Qt::Key_Super_R) {
             keyboard_input(1, event->key() == Qt::Key_Super_L ? 0x15B : 0x15C);
         } else
 #ifdef Q_OS_MACOS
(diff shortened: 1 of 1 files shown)


src/qt/qt_mainwindow.cpp
```diff
@@ -1646,6 +1652,9 @@ void MainWindow::blitToWidget(int x, int y, int w, int h, int monitor_index)
 
 void MainWindow::keyReleaseEvent(QKeyEvent* event)
 {
+    if (event->key() == Qt::Key_Pause) {
+        plat_pause(dopause ^ 1);
```
lemondrops, line 1656: Wouldn't this prevent the use of the Pause key inside the emulated machine?  
(Although this used to be a problem in the Win32 UI as well...)

Cacodemon345: Pause key isn't emulated.

lemondrops: What do you mean? It works for me.

Cacodemon345: Maybe only on Windows, but not on anything else.

lemondrops: Then shouldn't *that* be considered a bug, and fixed?

Cacodemon345: More like the pause key should perform only a single function at a time, not both.

lemondrops: Then the key to pause the emulator should be (and should have been ever since a long time ago) reassigned to a different key combination.

Cacodemon345: We would need to figure out one then.

Diicorp95: > Then the key to pause the emulator should be (and should have been ever since a long time ago) reassigned to a different key combination.

This could be an option (switch).

lemondrops (approved):

src/qt/qt_mainwindow.cpp
OBattler: I'm going to merge this now because it fixes an open issue. The pause button behavior can be fixed in the future.

PR #2648, opened 2022-08-27
Network improvements

elyosh:
Summary
=======
- Improve throughput to reach higher than 10Mbps speed (around 70 Mbps in my testing)
- Actually support other speeds than 10Mbps (currently with PCnet-FAST III)
- Add a feature that allows to disconnect / reconnect the link from the status bar icons

Checklist
=========
* [ ] Closes #xxx

References
==========


src/qt/qt_mediamenu.cpp
```diff
@@ -156,6 +160,16 @@ void MediaMenu::refresh(QMenu *parentMenu) {
         moMenus[i] = menu;
         moUpdateMenu(i);
     });
+
+    netMenus.clear();
+    MachineStatus::iterateNIC([this, parentMenu](int i) {
+        auto *menu = parentMenu->addMenu("");
+        netDisconnPos = menu->children().count();
+        auto *action = menu->addAction(tr("&Disconnected"), [i] { network_connect(i, !network_is_connected(i)); config_save(); });
```
cold-brewed, line 168:
```suggestion
        auto *action = menu->addAction(tr("&Connected"), [i] { network_connect(i, !network_is_connected(i)); config_save(); });
```
Small suggestion: I think it may be less confusing to show the state as "Connected" with a check vs. "Disconnected" with no check. In the default connected state it shows Disconnected without a check, but if you were not aware it can be enabled / disabled it makes it appear as if the current state is Disconnected.

src/qt/qt_mediamenu.cpp
```diff
@@ -662,6 +676,30 @@ void MediaMenu::moUpdateMenu(int i) {
...
+            netType = "PCAP";
+            break;
+    }
+
+    QString devName = DeviceConfig::DeviceName(network_card_getdevice(net_cards_conf[i].device_num), network_card_get_internal_name(net_cards_conf[i].device_num), 1);
+
+    auto *menu = netMenus[i];
+    auto childs = menu->children();
+    auto *connectedAction = dynamic_cast<QAction*>(childs[netDisconnPos]);
+    connectedAction->setChecked(!network_is_connected(i));
```
cold-brewed, line 698:
```suggestion
    connectedAction->setChecked(network_is_connected(i));
```
This is just to support the above suggestion of defaulting to "Connected" and enabled in the menu.

PR #2652, opened 2022-08-30
Add MT-32 (New) and CM-32LN

jriwanek:
Summary
=======
Adds MT-32 (New) and CM-32LN

Checklist
=========
* [ ] Closes #xxx
  * [X] I have opened a roms pull request - https://github.com/86Box/roms/pull/178/

References
==========
None


lemondrops:
When the device BIOS selector gets implemented, could the different ROM versions be moved there, instead of having separate devices?

lemondrops:
@jriwanek When the device BIOS selector gets implemented, could the different ROM versions be eventually consolidated there, instead of having separate devices for every ROM?

PR #2670, opened 2022-09-08
qt: Fix searching for icon packs in roms dir

cold-brewed:
Summary
=======
The existing code essentially hardcodes the icons path to `exe_path/roms`. Therefore the icon packs will only display in settings *if* the roms dir is in the same dir as the executable. With any other configuration (such as roms installed to global dir) the icon packs won't display.

This PR changes it to walk `rom_paths` and search each directory for `icons/` to find packs. Tested on mac, windows, linux and makes the icons appear for me in preferences.

Checklist
=========

References
==========
N/A

### DIFF
--- a/src/qt/qt_progsettings.cpp
+++ b/src/qt/qt_progsettings.cpp
@@ -46,16 +46,12 @@ QTranslator* ProgSettings::qtTranslator = nullptr;
 QString ProgSettings::getIconSetPath()
 {
     if (iconset_to_qt.isEmpty()) {
-        QVector<QFileInfo> roms_dirs;
-        // Walk rom_paths to get the candidates
-        for (rom_path_t *emu_rom_path = &rom_paths; emu_rom_path != nullptr; emu_rom_path = emu_rom_path->next) {
-            roms_dirs.append(QFileInfo(emu_rom_path->path));
-        }
         // Always include default bundled icons
         iconset_to_qt.insert("", ":/settings/win/icons");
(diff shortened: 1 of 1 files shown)


src/qt/qt_progsettings.cpp
```diff
@@ -43,24 +45,25 @@ ProgSettings::CustomTranslator* ProgSettings::translator = nullptr;
...
-    }
-
-    if (iconset_to_qt.isEmpty())
-    {
+    if (iconset_to_qt.isEmpty()) {
+        QVector<QFileInfo> roms_dirs;
+        // Walk rom_paths to get the candidates
+        for (rom_path_t *emu_rom_path = &rom_paths; emu_rom_path != nullptr; emu_rom_path = emu_rom_path->next) {
+            roms_dirs.append(QFileInfo(emu_rom_path->path));
+        }
```
dhrdlicka, line 53: Is there any particular reason why you loop twice over the same elements?

cold-brewed: Now that you mention it, hmmm... 😄 

I guess my brain somehow separated the tasks into 1) get the dirs then 2) search each for icons. 

In case it makes a difference, this code block is only executed once per program run because `iconset_to_qt.isEmpty()` will return false after the first run.

It does seem to make more sense to combine the two.

cold-brewed: I combined the two into a single loop.

PR #2755, opened 2022-10-27
Correct many file headers to show 86box

jriwanek:
Summary
=======
Correct many file headers to show 86box
Also fixes tabs to spaces in said headers (More to come in another PR)

Checklist
=========
* [ ] Closes #xxx

References
==========
None

### DIFF
--- a/src/codegen/codegen.h
+++ b/src/codegen/codegen.h
@@ -1,20 +1,20 @@
 /*
- * VARCem	Virtual ARchaeological Computer EMulator.
- *		An emulator of (mostly) x86-based PC systems and devices,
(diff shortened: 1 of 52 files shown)


dhrdlicka:
Pretty sure the printer stuff should be left alone, that was ported verbatim from VARCem.

OBattler:
> Pretty sure the printer stuff should be left alone, that was ported verbatim from VARCem.

I agree, also because that's code that uses the BSD 3-clause license, which explicitly mandates that the headers be left intact.

jriwanek:
Will fix as per discussion with Kado

PR #2758, opened 2022-10-28
qt: Update ui pause state in plat_pause

cold-brewed:
Summary
=======
This moves the ui pause button updates into `plat_pause`. A new function had to be added in `qt_mainwindow.cpp` to actually place the logic and updates due to `MainWindow:ui` being private.

This allows it to catch all `plat_pause` calls instead of just ones triggered from ui actions.

Checklist
=========

References
==========
N/A

### DIFF
--- a/src/qt/qt_mainwindow.hpp
+++ b/src/qt/qt_mainwindow.hpp
@@ -33,6 +33,7 @@ public:
     void blitToWidget(int x, int y, int w, int h, int monitor_index);
     QSize getRenderWidgetSize();
     void setSendKeyboardInput(bool enabled);
+    void setUiPauseState(bool paused);
 
     std::array<std::unique_ptr<RendererStack>, 8> renderers;
 signals:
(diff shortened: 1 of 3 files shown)


src/qt/qt_mainwindow.hpp
```diff
@@ -33,6 +33,7 @@ class MainWindow : public QMainWindow
     void blitToWidget(int x, int y, int w, int h, int monitor_index);
     QSize getRenderWidgetSize();
     void setSendKeyboardInput(bool enabled);
+    void setUiPauseState(bool paused);
```
Cacodemon345, line 36: This should be made a signal since Qt dislikes UI modifications from non-UI threads.

cold-brewed: Ah, did not realize it was a different thread. Thanks. I'll redo as a signal (I now see how similar signals are implemented elsewhere in `qt_platform.cpp` and called with `QTimer::singleShot`).

PR #2805, opened 2022-11-04
Modernize MSYS2 Makefile build

jriwanek:
Summary
=======
Modernize MSYS2 Makefile build so it uses similar names to the cmake build and actually produces artifacts too.

Checklist
=========
* [ ] Closes #xxx

References
==========
None


lemondrops:
Considering that using CMake has been the standard way to build 86Box for more than a year, do we still need this workflow? (Not the Makefile, mind you, but just the workflow)

PR #3084, opened 2023-02-05
qt: Various fields should not be editable in settings view

cold-brewed:
Summary
=======
When viewing the hard drive listing in the settings view, you can double click some of the fields and edit them. This disables the editable aspect of those fields (they shouldn't be editable).

Edit: Added the same fix for floppy, cd-rom, and other devices after @lemondrops brought it to my attention.

Checklist
=========

References
==========
N/A

### DIFF
--- a/src/qt/qt_settingsfloppycdrom.ui
+++ b/src/qt/qt_settingsfloppycdrom.ui
@@ -35,6 +35,9 @@
    </item>
    <item>
     <widget class="QTableView" name="tableViewFloppy">
+     <property name="editTriggers">
+      <set>QAbstractItemView::NoEditTriggers</set>
+     </property>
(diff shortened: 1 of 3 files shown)


lemondrops:
Can this fix be extended to other table views (floppy, CD-ROM, etc.) in the settings dialog? They seem to exhibit the same problem.

cold-brewed:
Good catch, I didn't even check the others! I'm also seeing it in:

* Floppy & CD-ROM
* Other removable

Are you seeing it anywhere else in settings?

lemondrops:
No, it's just the pages you mentioned.

cold-brewed:
Thanks, I added those as well.

PR #3085, opened 2023-02-05
qt: Fix editing fields at Hard Disks section

Cacodemon345:
Summary
=======
qt: Fix editing fields at Hard Disks section

Fields that can be changed by combo box selections are made not directly editable. CHS and Size values are made editable properly.

Checklist
=========

References
==========
None.

### DIFF
--- a/src/qt/qt_settingsharddisks.cpp
+++ b/src/qt/qt_settingsharddisks.cpp
@@ -65,6 +65,34 @@ normalize_hd_list()
 }
 */
 
+class HDDViewModel : public QStandardItemModel
+{
+    using QStandardItemModel::QStandardItemModel;
+
+public:
+    bool ignoreSetData = false;
+
+    bool setData(const QModelIndex &index, const QVariant &value, int role = Qt::EditRole) override
+    {
+        bool res = QStandardItemModel::setData(index, value, role);
+        if (ignoreSetData) return res;
+        if (index.column() == ColumnSize) {
+            uint32_t c = 0, h = 0, s = 0;
+            hdd_image_calc_chs(&c, &h, &s, value.toUInt());
+            QStandardItemModel::setData(this->index(index.row(), ColumnCylinders), c);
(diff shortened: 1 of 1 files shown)


lemondrops:
As I said in https://github.com/86Box/86Box/pull/3084#issuecomment-1418127309, the tables in other pages (floppy and CD-ROM, other removables) also have all their fields editable. Unlike the hard disks one, all the fields can be changed by combo boxes, though.

cold-brewed:
I have a few concerns with making the hdd field editable, at least in its current state.

First, the interface wasn't really designed with this in mind and would need some updates. See screenshot:
 
hdd edit

Due to the column size it's difficult / impossible to see the values being edited. The interface would need some updating and testing across platforms to avoid issues like above.

Second: Changing the CHS or size is a potentially *destructive* action on an existing drive from the perspective of the guest. As this is written, it just accepts any CHS / size modifications without any warnings. I feel like if you're going to enable this functionality there should at least be some warnings or guardrails involved. 

By the time someone updates the CHS / size with no warnings and restarts the guest it may be too late.

dhrdlicka:
> Changing the CHS or size is a potentially _destructive_ action on an existing drive from the perspective of the guest. As this is written, it just accepts any CHS / size modifications without any warnings. I feel like if you're going to enable this functionality there should at least be some warnings or guardrails involved.
> 
> By the time someone updates the CHS / size with no warnings and restarts the guest it may be too late.

Personally I think this shouldn't even be exposed from the UI at all. If a user needs to change the CHS of an existing drive, they can always edit the configuration file manually.

Dizzy611:
Hard pass on this idea. There is almost no reason for a user to want to edit the CHS on an existing drive and if they think they need to then we've got an X-Y problem going on. 

Suggest this does *NOT* get merged.

Cacodemon345:
I would be considering just exposing size editing (which makes more sense) but a proper 1:1 roundtrip CHS<->MiB conversion is impossible with our current code.

PR #3111, opened 2023-02-15
serial_passthrough: Allow configuration of named pipe parameter

Cacodemon345:
Summary
=======
serial_passthrough: Allow configuration of named pipe parameter

Windows-only.

Checklist
=========

References
==========
None.

### DIFF
--- a/src/include/86box/serial_passthrough.h
+++ b/src/include/86box/serial_passthrough.h
@@ -50,6 +50,7 @@ typedef struct serial_passthrough_s {
     intptr_t                     master_fd;    /* file desc for master pseudo terminal or
                                                 * socket or alike */
     char  host_serial_path[1024];              /* Path to TTY/host serial port on the host */
+    char  named_pipe[1024];                    /* (Windows only) Name of the pipe. */
     void *backend_priv;                        /* Private platform backend data */
 } serial_passthrough_t;
 
--- a/src/win/win_serial_passthrough.c
+++ b/src/win/win_serial_passthrough.c
@@ -22,6 +22,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <stdint.h>
+#include <wchar.h>
 
 #include <86box/86box.h>
 #include <86box/log.h>
@@ -30,6 +31,7 @@
 #include <86box/device.h>
 #include <86box/serial_passthrough.h>
 #include <86box/plat_serial_passthrough.h>
+#include <86box/ui.h>
 
 #include <windows.h>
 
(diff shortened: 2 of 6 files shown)


src/include/86box/serial_passthrough.h
```diff
@@ -50,6 +50,7 @@ typedef struct serial_passthrough_s {
     intptr_t                     master_fd;    /* file desc for master pseudo terminal or
                                                 * socket or alike */
     char  host_serial_path[1024];              /* Path to TTY/host serial port on the host */
+    char  named_pipe[1024];                    /* (Windows only) Name of the pipe. Prefixed with "\\.\86Box\" on the backend side. */
```
dhrdlicka, line 53: Why force a prefix in the first place? Neither VirtualBox nor VMware do that.

dhrdlicka (changes requested):

src/win/win_serial_passthrough.c
```diff
@@ -161,7 +161,8 @@ static int
 open_pseudo_terminal(serial_passthrough_t *dev)
 {
     char ascii_pipe_name[1024] = { 0 };
-    snprintf(ascii_pipe_name, sizeof(ascii_pipe_name), "\\\\.\\pipe\\86Box\\%s", vm_name);
+    if (dev->named_pipe[0] != 0) strncpy(ascii_pipe_name, dev->named_pipe, 1023);
+    else snprintf(ascii_pipe_name, sizeof(ascii_pipe_name), "\\\\.\\pipe\\86Box\\%s", vm_name);
```
dhrdlicka, line 165: Personally I think the user should have to type out the full path at all times.

Adding a non-obvious prefix if it's not a full path is pretty confusing. We should just error out if the provided string is not a valid named pipe path.

src/win/win_serial_passthrough.c
```diff
@@ -161,7 +162,15 @@ static int
...
-    snprintf(ascii_pipe_name, sizeof(ascii_pipe_name), "\\\\.\\pipe\\86Box\\%s", vm_name);
+    if (dev->named_pipe[0] != 0) strncpy(ascii_pipe_name, dev->named_pipe, 1023);
+    else {
+        ui_msgbox(MBX_WARNING, L"Serial Passthrough: Pipe name is empty.");
+        return 0;
+    }
+    if (strnicmp("\\\\.\\pipe\\", ascii_pipe_name, sizeof("\\\\.\\pipe\\") - 1)) {
+        ui_msgbox(MBX_WARNING, L"Serial Passthrough: Pipe name is invalid.");
+        return 0;
+    }
```
dhrdlicka, line 173: IMHO this condition is redundant as long as we properly handle `CreateNamedPipe` errors.

src/win/win_serial_passthrough.c
```diff
@@ -161,7 +162,15 @@ static int
...
     dev->master_fd = (intptr_t) CreateNamedPipeA(ascii_pipe_name, PIPE_ACCESS_DUPLEX, PIPE_TYPE_BYTE | PIPE_READMODE_BYTE | PIPE_NOWAIT, 32, 65536, 65536, NMPWAIT_USE_DEFAULT_WAIT, NULL);
     if (dev->master_fd == (intptr_t) INVALID_HANDLE_VALUE) {
         return 0;
```
dhrdlicka, line 176: I think we should show a message box here with the error information obtained via `GetLastError`.

PR #3250, opened 2023-04-12
macOS: Bundle fluidsynth and ghostscript

cold-brewed:
Summary
=======
This PR makes some changes to the build system to ensure fluidsynth and ghostscript get included in the macOS app bundle.

Nothing is changed regarding the functionality in 86box. As it is right now ghostscript and fluidsynth are supposed to be loaded at runtime on an as-needed basis.

However, this hasn't been working properly for some time. The libraries aren't bundled, requiring the user to have the library already installed (and even then, most of the time that will only work for *local* compiles). Manually placing the library in the app bundle generally does work but is inconvenient. 

To summarize: (edited)

* Updated cmake files to detect ghostscript and fluidsynth
* If detected, copy the library to the app bundle and fix up the `LC_ID_DYLIB`, `RPATH`, etc.
* ~Updated cmake files to find and link fluidsynth. This ensures the library gets copied to the bundle.~ edit: removed this. fluid is now treated the same as ghostscript - not linked and only copied if available.
* Some other changes to the qt `CMakeLists.txt` - renamed `QT_PLUGINS` to `RUNTIME_PLUGINS` since the var is now used for plugins in general, not just qt
* Add the packages to `dependencies_macports.txt` to make sure they get installed on the build server
* ~Fixed the fluid library name to `libfluidsynth.3.dylib` for loading~ Changed it back to `libfluidsynth.dylib` since it is being copied the same as ghostscript now

cc @richardg867 for review since this involves the external build system.

With this PR my local builds are fine (and even correct now) and builds on a test vm with ports seemed to work as well.

(further edit) **Note:** Fluid and / or Ghostscript will *only* get copied and bundled if the libraries are detected at build time. `cmake` will warn if they are not detected but will continue.

Checklist
=========

References
==========
N/A


cold-brewed:
Investigating the build failures. It may be due to fluidsynth not being installed as a dependency in GHA.

cold-brewed:
Build issues have been addressed so this should be good to go.

PR #3285, opened 2023-04-29
Initial softfloat port from Bochs to 86box, currently selectable only on QT.

TC1995:
Summary
=======
This should fix the 80-bit mess in the current code at the cost of performance, that's why it's an option.
Also fixes fonts in NeXTSTEP/OPENSTEP, IBM PS/2 model 70 type 4 POST 12903 error.
Passes all conformance tests of MCPDIAG as well as 87test, both for DOS.
Interpreter only, not for dynarec due to the way the latter is structured.

Checklist
=========
* [ ] Closes #xxx

References
==========
[Bochs source](https://github.com/bochs-emu/Bochs/tree/master/bochs/cpu)


dhrdlicka (changes requested):
The `x87_*` files should be moved out of the `softfloat` directory, since they are not a part of the actual softfloat library.

TC1995:
> The `x87_*` files should be moved out of the `softfloat` directory, since they are not a part of the actual softfloat library.

done

PR #3326, opened 2023-05-08
network: Add null network driver

cold-brewed:
Summary
=======
Adds a null network driver which replaces the previous type of `None`. TX packets are simply ignored. 

Previously, devices of type `None` were not added to the system. Now the devices will be added. There are UI updates to reflect the changes.

This also adds fallback to the null driver. If a different mode (PCAP, slirp, VDE) fails initialization it will fall back to the null driver.

## Details

`network.h`

Changed some comments to reflect new use of `NET_TYPE_NONE`
Fixed up a magic number in `_netcard_t`: Now uses `NET_QUEUE_COUNT`. This magic number was used in a few places in `network.c`.

`net_null.c`

The new driver. Any TX packets are ignored. To see the packets being tossed, define `ENABLE_NET_NULL_LOG`. RX packets are *not* polled because they are not expected nor received (as opposed to slirp or pcap where the respective descriptors must be polled).

`network.c`

Network cards will now default to Null Driver when no other modes are selected (PCAP, etc.). Previously, this was the mode of `None` in the network config. When using the null driver the cable is always in a disconnected state by default, regardless of the previous state.

In addition, when a particular network mode fails initialization, the system will fall back to the null driver rather than leave a null pointer behind, crashing the system. Initialization can fail for a variety of reasons, such as PCAP permissions. 

A temporary (untranslated) message appears to indicate that initialization failed and it will be falling back to null driver. This is temporary because an upcoming update to the network drivers will allow for obtaining the error message about *why* the init failed.

If the null driver init fails, that will result in a `fatal()` because the null driver shouldn’t fail. 

`network_dev_available()` no longer ignores `NET_TYPE_NONE`. Previously this would cause network cards of `None` type to not be added to the system. 

### QT

Places that previously showed `None` were updated to `Null Driver` (config settings, status bar popup). The network settings dialog was also updated to allow driver device selection when null driver is selected.

 (Note: There’s another small fix for how VDE was placed in the logic for device selection)

### Win32

I did my best here to update the network settings dialog to now show `Null Driver` instead of `None`.  I didn’t, however, update the logic to allow network device selection when null driver is selected. You’ll have to temporary change to a different mode (like slirp), change the device, then change back to null driver.

Checklist
=========

References
==========
N/A


cold-brewed:
I forgot to commit the updated `Makefile.mingw`, now added.

PR #3365, opened 2023-05-23
Update network drivers to provide error message

cold-brewed:
Summary
=======
Continuation of some work done in #3326 

This PR modifies the network drivers to have `netdrv_errbuf` as part of the init. As part of normal operation, when the init fails it returns `NULL` and there is no way to understand *why* the init failed. Now, when an init fails an error message with more information is copied to `netdrv_errbuf` so the UI can provide some information to the user.

All of the existing drivers were updated to use this new functionality and provide the error message (slirp, pcap, vde).

New translation strings were added to qt and win32 as well. Upon failure, the error is displayed and network falls back to the null driver. @jriwanek might want to review and make sure I didn't break win32 (though it does compile and run fine in my testing).

A separate issue was fixed in the network configuration dialog with VDE where it was grabbing config data incorrectly and getting corrupt data.

Tested on mac, windows and linux. 

Checklist
=========

References
==========
#3326

### DIFF
--- a/src/include/86box/language.h
+++ b/src/include/86box/language.h
@@ -163,6 +163,8 @@
 #    define IDS_DYNAREC IDS_2163
 #endif
 #define IDS_2166      2166 // "Video card #2 ""%hs"" is not..."
+#define IDS_2167      2167 // "Network driver initialization failed"
+#define IDS_2168      2168 // "The network configuration will be switched to the null driver"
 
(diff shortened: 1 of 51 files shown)


cold-brewed:
The CI failures appear to be from an issue with sonar failing to run.. not with the code itself.

PR #3393, opened 2023-06-10
Add some defines to build succesfully on FreeBSD

chungy:
Tested on FreeBSD 13.2 only.  FreeBSD is more similar to Mac OS X in system-level library support than it is to Linux.

So far as I know, need to use `CFLAGS=-I/usr/local/include/freetype2 cmake ...` to get cmake to include the proper FreeType 2 headers. If there's a cmake expert that could fix it, that'd be great.

### DIFF
--- a/src/include/86box/plat_dir.h
+++ b/src/include/86box/plat_dir.h
@@ -66,6 +66,9 @@ extern void           seekdir(DIR *, long);
 extern int            closedir(DIR *);
 
 #    define rewinddir(dirp) seekdir(dirp, 0L)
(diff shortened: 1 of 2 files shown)


waltje:
Geez, can these be disabled ?


    On Saturday, June 10, 2023 at 07:40:45 PM EDT, sonarcloud[bot] ***@***.***> wrote:  
 
 


Kudos, SonarCloud Quality Gate passed!    

  0 Bugs
  0 Vulnerabilities
  0 Security Hotspots
  0 Code Smells

 No Coverage information
 0.0% Duplication

PR #3397, opened 2023-06-11
qt: Hide the VDE socket textbox on Windows

lemondrops:
Summary
=======
Title.
Also fixes an unused variable warning as a side effect.

Checklist
=========


lemondrops:
@jriwanek Why does the sonarcloud bot have to post its message *eight times*? Sure, it deletes the older ones, but it's still quite spammy with notifications on:
image

PR #3486, opened 2023-07-23
Actually fix #3484

### LINKED ISSUE #3484: Incorrect fields width in the Network Card tab 3
**Describe the bug**
In the Network Card 3 tab there is an incorrect fields width: Mode, Interface, Adapter, VDE Socket (it differs from the others). In all other tabs: Network Card 1, Network Card 2, Network Card 4, the width of these fields is the same.

**Expected behavior**
Fields width: Mode, Interface, Adapter, VDE Socket must be the same in all Network Card tabs.

 - OS: Windows 10 Pro x64 v22H2
 - 86Box v4.0.0 x86 Build 4724
 - Build type: qt, regular, old recompiler

lemondrops:
Summary
=======
An actual fix for #3484 by means of removing a trailing space in a label that caused the controls to shift slightly to the right. Reverts the previous failed attempt just to be safe.

Screenshot 2023-07-23 162830
Screenshot 2023-07-23 162916

Checklist
=========
* [x] Closes #3484

### DIFF
--- a/src/qt/qt_settingsnetwork.ui
+++ b/src/qt/qt_settingsnetwork.ui
@@ -36,52 +36,6 @@
        <string>Network Card #1</string>
       </attribute>
       <layout class="QGridLayout" name="gridLayout_3">
-       <item row="0" column="0">
-        <widget class="QLabel" name="label">
-         <property name="sizePolicy">
-          <sizepolicy hsizetype="Minimum" vsizetype="Preferred">
-           <horstretch>0</horstretch>
-           <verstretch>0</verstretch>
-          </sizepolicy>
-         </property>
-         <property name="text">
-          <string>Mode</string>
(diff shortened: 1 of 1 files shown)


OBattler:
Can this be fixed without reordering al the items in the .ui file back to the previous mess?

lemondrops:
How is it a mess?

OBattler:
- lemondrops: If you open the .ui file in something like Notepad++, you'll see that the items are in a random order. But, I was now told that this is normal for Designer, so I have decided not to pursue it further.

lemondrops:
Weird, it didn't look random to me when I looked at it. But eh, whatever, I guess. And yes, Qt Designer will rearrange them anyway, which is rather annoying as it makes diffs harder to read.

PR #3543, opened 2023-08-11
Video changes: (see below)

TC1995:
Summary
=======
1. The passthrough from VGA to 8514/A and/or 8514/A to VGA no longer relies on hackish places where to switch from/to, instead, relying on port 0x3c3 of VGA doing so (though the Mach8/32 still needs some places where to manually switch from/to, mainly the MCA one when configuring the EEPROM).
2. Implemented the MCA behalf of the Mach32 and its corresponding reset function as well as its ROM.
3. Properly implemented (more or less) true color, including 24-bit BGR rendering
4. Other fixes such as color patterns and mono patterns being more correct than before in various operating systems and in 24-bit true color.
5. Implemented the onboard Mach32 video of the IBM PS/ValuePoint P60 machine.
6. Made the onboard internal video detect when it's 8514/A compatible or not (CGA/EGA/MDA/VGA/etc.). If the former is selected, then the video monitor flag is used instead (for QT).
7. The TGUI9400 and 9440, if on VLB, now detect the right amount of memory if on 2MB.
8. Initial implementation of the ATI 68875 ramdac used by the Mach32 and made the ATI 68860 8514/A aware when selected with the Mach32AX PCI.
9. Separated the 8514/A ramdac ports from the VGA ramdac ports, allowing seamless transition from/to 8514/A/VGA.
10. Fixed a hdisp problem in the ET4000/W32 cards, where it was doubling the horizontal display in 15bpp+ graphics mode.
11. Removed the 0x3da/0x3ba port hack that was on the Mach8/32 code, relying on the (S)VGA core instead.
12. Reworked and simplified the TGUI9440 pitch register based on logging due to no documentation at all.
13. Used a newer PCI Mach32 BIOS ROM and another BIOS ROM of Mach32 ISA which says "ATI Ultra Pro".

Checklist
=========
* [ ] Closes #xxx
  * [x] I have opened a roms pull request - https://github.com/86Box/roms/pull/206/

References
==========

### DIFF
--- a/src/include/86box/vid_8514a.h
+++ b/src/include/86box/vid_8514a.h
@@ -19,7 +19,16 @@
 #ifndef VIDEO_8514A_H
 #define VIDEO_8514A_H
 
+typedef struct {
+    int ena,
+        x, y, xoff, yoff, cur_xsize, cur_ysize,
+        v_acc, h_acc;
+    uint32_t addr, pitch;
+} hwcursor8514_t;
+
 typedef struct ibm8514_t {
+	hwcursor8514_t hwcursor;
+	hwcursor8514_t hwcursor_latch;
     uint8_t pos_regs[8];
 
     int force_old_addr;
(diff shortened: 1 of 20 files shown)


lemondrops:
You don't have to put (see below) in the title, that's already obvious.

PR #3681, opened 2023-09-05
Allow connection to a VDE switch run by the same user

redoste:
Summary
=======
Allow connection to a VDE switch run by the same user because `libvdeplug` will only change the mode if the switch is run by another user that isn't root.
If the switch is run by the same user as 86Box the socket will be chmoded to 000 and the switch, unable to connect back, will drop the connection.

Checklist
=========
* [ ] Closes #xxx

References
==========
https://github.com/rd235/vdeplug4/blob/187256c528137036cc4e3228728b3f70a3342fa5/libvdeplug4/libvdeplug_vde.c#L261-L275


cold-brewed:
Thanks for the PR!

Issue #3224, opened 2023-03-24
Serial passthrough not working properly


mike-cuche:
I'm trying to use the serial passthrough. I set it up via the settings window, and made a basic loopback by shorting pins 2 and 3 on the port, however it does not work (it does on the host, Dosbox-x and Qemu VMs, though.) and I have no clue on how to debug this

This is my 86box.cfg; the guest OS is Windows 98 SE, tests done with Putty and HyperTerminal.

The baud rate I had set is the maximum supported by my serial port.
```
[General]
vid_renderer = qt_software
scale = 3
window_remember = 1
video_filter_method = 0
video_fullscreen_scale = 3
confirm_exit = 0
dpi_scale = 0
language = system
hide_tool_bar = 1

[Machine]
machine = borapro
cpu_family = pentium2_klamath
cpu_speed = 266666666
cpu_multi = 4
cpu_use_dynarec = 1
time_sync = local
fpu_type = internal
mem_size = 131072

[Video]
gfxcard = voodoo3_3k_pci

[Input devices]
mouse_type = ps2
joystick_type = 4axis_4button
joystick_0_nr = 1
joystick_0_axis_0 = 0
joystick_0_axis_1 = 1
joystick_0_axis_2 = 3
joystick_0_axis_3 = 4
joystick_0_button_0 = 0
joystick_0_button_1 = 1
joystick_0_button_2 = 2
joystick_0_button_3 = 3

[Storage controllers]
hdc = internal
cassette_mode = load

[Standard PS/2 Mouse]
buttons = 4

[3dfx Voodoo3 3000]
bilinear = 1
dithersub = 1
dacfilter = 0
render_threads = 2
recompiler = 1

[Ports (COM & LPT)]
serial2_enabled = 0
serial1_passthrough_enabled = 1

[Floppy and CD-ROM drives]
fdd_01_type = 35_2ed
fdd_02_type = none
cdrom_01_parameters = 1, atapi
cdrom_01_ide_channel = 1:0
cdrom_01_speed = 72
cdrom_01_type = 86BOX_CD-ROM_1.00

[Sound Blaster 16]
base = 0220
base401 = 0000
irq = 5
dma = 1
dma16 = 5
opl = 1
receive_input = 1
receive_input401 = 1

[FluidSynth]
sound_font = weedsgm4.sf2
output_gain = 20
chorus = 0
chorus_voices = 2
chorus_level = 100
chorus_speed = 30
chorus_depth = 80
chorus_waveform = 0
[...]

PR #3691, opened 2023-09-15
fix flags for serial passthrough on unix

adh:
Summary
=======
The logic for clearing bits in termios.c_flag was reversed. That caused clearing of all the other flags than which were intended to be cleared, which in turn caused the host serial port to be stuck at random and unusable configuration (5 data bits on macOS).

Checklist
=========
* [ ] Closes #xxx

### DIFF
--- a/src/unix/unix_serial_passthrough.c
+++ b/src/unix/unix_serial_passthrough.c
@@ -182,7 +182,6 @@ plat_serpt_set_params(void *priv)
                 term_attr.c_cflag |= CMSPAR;
 #endif
         }
-        printf("c_cflag=%08x\n", term_attr.c_cflag);
         tcsetattr(dev->master_fd, TCSANOW, &term_attr);
 #undef BAUDRATE_RANGE
     }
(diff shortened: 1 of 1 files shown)


cold-brewed:
Thank you for the PR. Did you notice the serial port being stuck in random configurations while serial passthrough was in use, or after exiting 86Box?

adh:
In use. The bug this fixes made the serial passthrough essentially unusable.

adh:
Seems to be related to https://github.com/86Box/86Box/issues/3224

cold-brewed:
Alright. I was able to previously (without the fix) access a serial console in a linux guest, macOS host, using `minicom`. Though I did notice some odd behavior which may be related. I'll test out this PR to see if it addresses the issues I had.

adh:
Note that this is relevant for passthrough to a real physical host serial port (tested on macOS host with FTDI serial converter).  It does not matter for ptys as the there is no hardware UART to configure.

Without the fix sending U (0x55) from the VM leads to this:
IMG_6941

ie. 5 databits, not 8 as in the configuration file. The reason for this behavior is that the expression on line 168 clears the configured number of databits.

PR #3725, opened 2023-10-06
Updated Simp. Chinese and Japanese translations

EmpyreusX:
Summary
=======
Updated Simp. Chinese and Japanese translations.


lemondrops:
Please also update `src/qt/languages/ja-JP.po` and `src/qt/languages/zh-CN.po`. These files are for the currently used Qt-based UI, while the .rc files are for the deprecated Win32 UI.

Issue #1951, opened 2021-12-21
SCO OpenServer 5.0.7 crashes when you try to start the GUI


mjkerpan:
**Describe the bug**
The X11 GUI subsystem in OpenServer crashes with the following error when I attempt to launch the GUI: "WARNING: u_trap - Invalid TSS 0x0000000A". I tried to run this with both a Trio64 and an ET4000/w32 selected.

**To Reproduce**
1. Login to an installed SCO OpenServer system
2. Run "startx" or set the system to boot to the graphical login system

**Expected behavior**
The GUI system should start and the desktop should appear.

**Screenshots**
![20211220-020426-517](https://user-images.githubusercontent.com/9254521/146861365-562f6932-4e7e-4a0f-926f-fa82f0ef6e7b.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - 86Box version: 3389
 - Build information: Default build

OBattler:
I think this can now be closed.

PR #3811, opened 2023-11-07
add cpl_override in pmodeiret() when cpu_state.flags has NT_FLAG

jgilje:
Summary
=======
I've been hit by #1951 myself, and debugging the issue led me to guess cpl_override should be set for this read. With this patch, the warnings disappear, but OpenServer still doesn't go graphical unless a very small subset of video cards has been selected.

Checklist
=========
* [ ] Closes #xxx

References
==========
Debugging the same CPU state in QEMU helped me verify the expected value for the `seg` assignment.

### DIFF
--- a/src/cpu/x86seg.c
+++ b/src/cpu/x86seg.c
@@ -1796,7 +1796,9 @@ pmodeiret(int is32)
     }
 
     if (cpu_state.flags & NT_FLAG) {
+        cpl_override = 1;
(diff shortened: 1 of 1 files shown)


jgilje:
With the second commit I added to this PR, this OS now boots graphically for all the cards I've tested (and they all failed to detect and run before this patch)

PR #3975, opened 2023-12-31
fixes typo on mac icon fix

brunocastello:
Summary
=======
This fixes a typo on a mac icon fix submitted by lemondrops.
The original code uses 
`#ifndef Q_OS_APPLE `
instead of 
`#ifndef __APPLE__`

After changing this line, the correct icon appears on MacOS Dock, for both app window and settings window.

Checklist
=========
* [X] Closes #3973 

References
==========
OBattler

### DIFF
--- a/src/qt/qt_main.cpp
+++ b/src/qt/qt_main.cpp
@@ -195,7 +195,7 @@ main(int argc, char *argv[])
     SetCurrentProcessExplicitAppUserModelID(L"86Box.86Box");
 #endif
 
-#ifndef __APPLE__
+#ifndef Q_OS_MACOS
 #    ifdef RELEASE_BUILD
     app.setWindowIcon(QIcon(":/settings/win/icons/86Box-green.ico"));
 #    elif defined ALPHA_BUILD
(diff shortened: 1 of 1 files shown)


lemondrops:
Are you sure? They should evaluate to the same thing.

src/qt/qt_main.cpp
```diff
@@ -195,7 +195,7 @@ main(int argc, char *argv[])
     SetCurrentProcessExplicitAppUserModelID(L"86Box.86Box");
 #endif
 
-#ifndef Q_OS_APPLE
+#ifndef __APPLE__
```
lemondrops, line 198:
```suggestion
#ifndef Q_OS_MACOS
```

lemondrops (changes requested):

src/qt/qt_main.cpp
brunocastello: I'll test it @lemondrops

brunocastello: @lemondrops yup, your proposed change works, I'll accept it, ty!

lemondrops: Thanks! I guess I should've double-checked the Qt documentation instead of trying to recall it from memory :)

lemondrops (approved):

PR #4031, opened 2024-01-12
Qt: Use "fusion" style when dark mode is activated on Windows 10/11

Cacodemon345:
Summary
=======
Qt: Use "fusion" style when dark mode is activated on Windows 10/11

Checklist
=========

References
==========
None.

### DIFF
--- a/src/qt/qt_main.cpp
+++ b/src/qt/qt_main.cpp
@@ -154,9 +154,45 @@ main_thread_fn()
 
 static std::thread *main_thread;
 
+#ifdef Q_OS_WINDOWS
+static bool is_light_theme() {
+    // based on https://stackoverflow.com/questions/51334674/how-to-detect-windows-10-light-dark-mode-in-win32-application
+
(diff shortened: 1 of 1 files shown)


cold-brewed:
There are a couple of problems with this approach.

First, the fusion style doesn't match the native OS look. Qt seems to have intended it to be a *fusion* of all the different styles.

Second, The dark mode support also won't match the OS native dark mode, ultimately resulting in different colors *and* different widget styles. This PR sets that by default when the OS dark mode is set which would be even more jarring.

You'd in effect have a third style that would not match standard light or dark mode. It would make the application stand out even more instead of blending in with the native dark mode.

Unfortunately Qt doesn't make it easy to support dark mode on windows. We're also still using Qt 5 by default.

A proper solution to this would be to go the route dolphin took with dark mode (which is much more involved but has a much better result). This should be put aside until we can pursue a proper solution.

Cacodemon345:
I don't know about what Dolphin does but trying to roll out custom dark mode support especially for Windows will not work out well at the end without serious commitment to make it not look atrocious.

PR #4061, opened 2024-01-17
Switch to libspng from libpng

Cacodemon345:
Summary
=======
Switch to libspng from libpng.

Also adds in miniz with it, which is said to be better than zlib according to official docs.

Checklist
=========

References
==========
None.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -74,7 +74,7 @@ if(INSTRUMENT)
 endif()
 
 target_link_libraries(86Box cpu chipset mch dev mem fdd game cdrom zip mo hdd
-    net print scsi sio snd vid voodoo plat ui)
+    net print scsi spng sio snd vid voodoo plat ui)
 
 if(WIN32 AND ARCH STREQUAL "i386")
     if(MINGW)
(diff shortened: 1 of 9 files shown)


dhrdlicka:
Are there any actual benefits that outweigh the fact this is another in-tree external dependency that we need to maintain? I thought we were trying to get rid of these.

Cacodemon345:
These are relatively trivial to maintain; the libraries are designed to be embedded into projects. The API is more sane than in libpng, and it will reduce binary size.

miniz in particular has a stable API.

cold-brewed:
As @dhrdlicka said, is there a compelling reason to switch to an in-tree dependency from an established, existing, standard external library?

What problem is this solving? Our png usage is minimal so I don't find the API argument very compelling. Whether trivial or not, it would still add another in-tree dependency to maintain.

Cacodemon345:
It's solving the problem of relying on a large dependency.

Most of the functionality of libpng isn't used in 86Box. The entirety of its decoding API is unused, and any encoding API usage is minimal.

With 86Box's only usage of libpng being solely for encoding, it's not wise to rely on it because it will account for everything supported in PNG, and it will generally bloat up debug builds.

Hence the switch. It's also sufficient and working well for our use cases and it doesn't need to be maintained due to its embeddable nature. The library is feature complete in its current state as well.

cold-brewed:
We don't distribute debug builds so that is a non-issue.

There are probably other libraries in use where we don't need or use the full functionality - that's the whole point of a library. I don't see that a reason to move to an embedded dependency of that library unless there is a specific issue or need.

PR #4092, opened 2024-01-28
Qt: MAC address configuration

Cacodemon345:
Summary
=======
Qt: MAC address configuration

Checklist
=========

References
==========
None.

### DIFF
--- a/src/device.c
+++ b/src/device.c
@@ -450,8 +450,7 @@ device_has_config(const device_t *dev)
     config = dev->config;
 
     while (config->type != -1) {
-        if (config->type != CONFIG_MAC)
-            c++;
+        c++;
         config++;
     }
 
--- a/src/qt/qt_deviceconfig.cpp
+++ b/src/qt/qt_deviceconfig.cpp
@@ -38,6 +38,7 @@ extern "C" {
 #include <86box/device.h>
 #include <86box/midi_rtmidi.h>
 #include <86box/mem.h>
+#include <86box/random.h>
 #include <86box/rom.h>
 }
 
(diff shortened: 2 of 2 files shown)


OBattler:
I'll have to manually port these changes since I have major device configuration changes on my end as I'm trying to add the possibility of BIOS-dependent configuration options (which are going to be needed for eg. graphics cards where different BIOS'es can mean eg. different minimum and maximum RAM amounts).

lemondrops:
Also, how does this behave when there's no MAC address generated yet (i.e. when a network card is newly added?)

Cacodemon345:
If the user doesn't open the configuration settings for the network device after selecting it, the MAC address is automatically generated. If the user does after selecting it (but before saving the settings), and he/she doesn't alter the default address of 00:00:00 in it and leaves it as is when confirming the settings, it will be saved as 00:00:00 and it will not be randomly generated when the emulated machine starts with this network device.

dhrdlicka:
This is likely out of scope but we might want to consider letting the user change the entire MAC address and not only the second half. Especially if we want to expose this as a UI option

OBattler:
> This might be out of scope but we should consider letting the user change the entire MAC address and not only the second half. Especially if we want to expose this as a UI option

Given that this can affect device detection (see NE2000), I don't tink that's a good idea.

dhrdlicka:
That raises the question whether there is any point in exposing this in the user interface in the first place if the full MAC address cannot be changed. I cannot think of a scenario where a user would only want to change the second half of the MAC address. That is, other than a MAC address conflict, which would be so rare that it makes more sense to just instruct users to remove the conflicting address from the config to generate a new one.

Most (if not all) usecases for letting the MAC address be configurable expect it to be configurable in its full length. Usually that would involve using PCap on some strange network setup where the router keeps track of client MAC addresses.

And sure enough, when an option to change the MAC address is exposed, it lets you change the full address:
image
image

lemondrops:
> Given that this can affect device detection (see NE2000), I don't tink that's a good idea.

Well, if a user overrides the vendor portion to something that affects device detection, it would be on them and not us, wouldn't it?

> And think of it this way - real cards of the same vendor would have the same first half of the MAC address and only differ in the second half.

Good thing 86Box is an emulator and not real hardware :D Also to be pedantic, the "same first half" part is only true if the vendor has only one OID, and many vendors have multiple.

dhrdlicka:
> Speaking of VMware, it automatically detects if a VM was moved from its previous location, and asks the user if they moved or copied it, so that it can generate a new MAC address if the latter option is selected, avoiding a conflict in the example case above. Maybe something like tht could be implemented in 86Box?

@lemondrops We could generate the second half of the MAC address as a hash of the VM path and check if they match on startup. If not, then ask the user whether the address should be regenerated. If they don't want to regenerate it, then save a boolean flag to the config to disable the check.

PR #4110, opened 2024-02-02
Device runtime configuration (initiallly for MIDI devices)

Cacodemon345:
Summary
=======
This PR introduces device runtime configuration support. Initially done for RtMidi and FluidSynth MIDI backends.

Checklist
=========

References
==========
None.


src/qt/qt_deviceconfig.cpp
```diff
@@ -365,6 +377,14 @@ DeviceConfig::ConfigureDevice(const _device_ *device, int instance, Settings *se
             }
             config++;
         }
+
+        /* The :: prefix is intentional. */
```
dhrdlicka, line 381: Maybe explain _why_ it's there rather than just say that it's intended?

src/qt/qt_mainwindow.ui
```diff
@@ -87,7 +87,14 @@
     <property name="title">
      <string>&amp;Tools</string>
     </property>
+    <widget class="QMenu" name="menuDevice_settings">
+     <property name="title">
+      <string>Device settings...</string>
```
dhrdlicka, line 92:
```suggestion
      <string>Device settings</string>
```
There should be no ellipsis at the end, since it's a submenu.

dhrdlicka (dismissed):

dhrdlicka:
I think ideally this should just be implemented in the Settings dialog, which would detect whether the selected devie config changes require a hard reset, but I guess this is a good start towards that.

src/qt/qt_deviceconfig.cpp
Cacodemon345: I clarified it. I think the explanation I gave is good enough.

PR #4270, opened 2024-03-15
Compile fixes

Cacodemon345:
Summary
=======
Compile fixes.

Win32 GUI builds now build properly.

Checklist
=========

References
==========
None.

### DIFF
--- a/src/qt/qt_main.cpp
+++ b/src/qt/qt_main.cpp
@@ -220,13 +220,13 @@ main(int argc, char *argv[])
 
 #ifdef Q_OS_WINDOWS
 #    if !defined(EMU_BUILD_NUM) || (EMU_BUILD_NUM != 5624)
-    HWND winbox = FindWindow("TWinBoxMain", NULL);
+    HWND winbox = FindWindowW(L"TWinBoxMain", NULL);
     if (winbox &&
-        FindWindowEx(winbox, NULL, "TToolBar", NULL) &&
-        FindWindowEx(winbox, NULL, "TListBox", NULL) &&
-        FindWindowEx(winbox, NULL, "TStatusBar", NULL) &&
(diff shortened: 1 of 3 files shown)


lemondrops:
Please explain what do changes in qt_main.cpp fix.

Cacodemon345:
It changes the functions to use explicit Unicode calls, to avoid compile errors in case of UNICODE define mismatch.

cold-brewed:
> It changes the functions to use explicit Unicode calls, to avoid compile errors in case of UNICODE being defined.

Do we even make use of such a configuration? I'm not seeing any related compilation errors in any of our build environments.

The changes being made to `qt_main.cpp` seem unrelated to modem emulation or the win32 ui. PRs should be limited in scope to the specific issue being addressed.

Also note that win32 and therefore makefile building is deprecated.

Cacodemon345:
And to add to my point, many Win32 UI files always define UNICODE. So far the current Qt UI haven't done so on Windows.

cold-brewed:
The PR title is sufficiently vague so just about anything could fall under "Compile Fixes."

The title also seems to imply that the fixes are required for compiling, and as far as I can tell qt was compiling without an issue. It's hard to tell when very little information is provided.

As for win32: It is currently only built as part of the CodeQL workflow. The fact that it is still enabled there is just an oversight from when we initially disabled win32 builds. That will be taken care of.

Side note: This type of information should be included in the initial PR text.

Cacodemon345:
I'd have noted this issue in the initial PR text if I was aware that this was related to Qt6.

cold-brewed:
This PR is no longer valid. GHA win32 has been fully disabled in master (#4273) and  the files removed from the tree in 4.2.

OBattler:
Closing this, the QT change can be re-PR'd in a new PR with a proper explanation.

PR #4274, opened 2024-03-15
Compile fixes for Qt6 on Windows

Cacodemon345:
Summary
=======
Qt6 on Windows defines the UNICODE macro by default unlike Qt5. This PR fixes compilation with Qt6 on Windows by forcing it and changing the WinBox manager checks to use Unicode strings.

Checklist
=========

References
==========
None.

### DIFF
--- a/src/qt/qt_main.cpp
+++ b/src/qt/qt_main.cpp
@@ -17,6 +17,10 @@
  *          Copyright 2021-2022 Cacodemon345
  *          Copyright 2021-2022 Teemu Korhonen
  */
+
+#ifdef _WIN32
+#define UNICODE 1
+#endif
 #include <QApplication>
 #include <QSurfaceFormat>
 #include <QDebug>
@@ -220,13 +224,13 @@ main(int argc, char *argv[])
 
 #ifdef Q_OS_WINDOWS
 #    if !defined(EMU_BUILD_NUM) || (EMU_BUILD_NUM != 5624)
-    HWND winbox = FindWindow("TWinBoxMain", NULL);
+    HWND winbox = FindWindow(L"TWinBoxMain", NULL);
     if (winbox &&
-        FindWindowEx(winbox, NULL, "TToolBar", NULL) &&
-        FindWindowEx(winbox, NULL, "TListBox", NULL) &&
-        FindWindowEx(winbox, NULL, "TStatusBar", NULL) &&
-        (winbox = FindWindowEx(winbox, NULL, "TPageControl", NULL)) && /* holds a TTabSheet even on VM pages */
-        FindWindowEx(winbox, NULL, "TTabSheet", NULL))
+        FindWindowEx(winbox, NULL, L"TToolBar", NULL) &&
+        FindWindowEx(winbox, NULL, L"TListBox", NULL) &&
+        FindWindowEx(winbox, NULL, L"TStatusBar", NULL) &&
+        (winbox = FindWindowEx(winbox, NULL, L"TPageControl", NULL)) && /* holds a TTabSheet even on VM pages */
+        FindWindowEx(winbox, NULL, L"TTabSheet", NULL))
 #    endif
     {
         QMessageBox warningbox(QMessageBox::Icon::Warning, QObject::tr("WinBox is no longer supported"),
--- a/src/qt/qt_main.cpp
+++ b/src/qt/qt_main.cpp
@@ -17,10 +17,6 @@
  *          Copyright 2021-2022 Cacodemon345
  *          Copyright 2021-2022 Teemu Korhonen
  */
-
-#ifdef _WIN32
-#define UNICODE 1
-#endif
 #include <QApplication>
 #include <QSurfaceFormat>
 #include <QDebug>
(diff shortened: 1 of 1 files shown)


lemondrops:
Don't we already define `UNICODE` in <86box/win.h>, which is included in qt_main.cpp?

Cacodemon345:
It is defined, yes, but I think UNICODE needs to be defined to `1` instead of an empty value.

cold-brewed:
> It is defined, yes, but I think UNICODE needs to be defined to `1` instead of an empty value.

Have you verified that it is required to be like that?

Cacodemon345:
Actually, I think it needs to be defined before including any headers. So including <86box/win.h> in the middle wouldn't work.

Also, on second thought, UNICODE can be defined to an empty value since that's how the Win32 UI did it.

Biswa96:
I have brewed a different patch. It is not required to define `_UNICODE` macro explicitly. In the following patch, the string data type (char vs. wchar_t) is automatically set with `_T()` macro.

<details><summary>Click here to see the diff</summary>

```diff
--- a/src/qt/qt_main.cpp
+++ b/src/qt/qt_main.cpp
@@ -46,6 +46,7 @@
 #    include "qt_winmanagerfilter.hpp"
 #    include <86box/win.h>
 #    include <shobjidl.h>
+#    include <tchar.h>
 #endif
 
 extern "C" {
@@ -220,13 +221,13 @@
 
 #ifdef Q_OS_WINDOWS
 #    if !defined(EMU_BUILD_NUM) || (EMU_BUILD_NUM != 5624)
-    HWND winbox = FindWindow("TWinBoxMain", NULL);
+    HWND winbox = FindWindow(_T("TWinBoxMain"), NULL);
     if (winbox &&
-        FindWindowEx(winbox, NULL, "TToolBar", NULL) &&
-        FindWindowEx(winbox, NULL, "TListBox", NULL) &&
-        FindWindowEx(winbox, NULL, "TStatusBar", NULL) &&
-        (winbox = FindWindowEx(winbox, NULL, "TPageControl", NULL)) && /* holds a TTabSheet even on VM pages */
-        FindWindowEx(winbox, NULL, "TTabSheet", NULL))
+        FindWindowEx(winbox, NULL, _T("TToolBar"), NULL) &&
+        FindWindowEx(winbox, NULL, _T("TListBox"), NULL) &&
+        FindWindowEx(winbox, NULL, _T("TStatusBar"), NULL) &&
+        (winbox = FindWindowEx(winbox, NULL, _T("TPageControl"), NULL)) && /* holds a TTabSheet even on VM pages */
+        FindWindowEx(winbox, NULL, _T("TTabSheet"), NULL))
 #    endif
     {
         QMessageBox warningbox(QMessageBox::Icon::Warning, QObject::tr("WinBox is no longer supported"),
```

</details>

Biswa96:
Little example about what it does.

* If `_UNICODE` is defined:
  - `FindWindow` becomes `FindWindowW`
  - `"TWinBoxMain"` becomes `L"TWinBoxMain"`
* If `_UNICODE` is not defined, no change.

src/qt/qt_main.cpp
```diff
@@ -17,6 +17,10 @@
  *          Copyright 2021-2022 Cacodemon345
  *          Copyright 2021-2022 Teemu Korhonen
  */
+
+#ifdef _WIN32
+#define UNICODE 1
+#endif
```
dhrdlicka, line 23: No need, just don't use the tchar macros and call the function with the appropriate suffix explicitly if you're hardcoding narrow/wide strings

src/qt/qt_main.cpp
```diff
@@ -220,13 +224,13 @@ main(int argc, char *argv[])
...
-        FindWindowEx(winbox, NULL, "TToolBar", NULL) &&
-        FindWindowEx(winbox, NULL, "TListBox", NULL) &&
-        FindWindowEx(winbox, NULL, "TStatusBar", NULL) &&
-        (winbox = FindWindowEx(winbox, NULL, "TPageControl", NULL)) && /* holds a TTabSheet even on VM pages */
-        FindWindowEx(winbox, NULL, "TTabSheet", NULL))
+        FindWindowEx(winbox, NULL, L"TToolBar", NULL) &&
+        FindWindowEx(winbox, NULL, L"TListBox", NULL) &&
+        FindWindowEx(winbox, NULL, L"TStatusBar", NULL) &&
+        (winbox = FindWindowEx(winbox, NULL, L"TPageControl", NULL)) && /* holds a TTabSheet even on VM pages */
+        FindWindowEx(winbox, NULL, L"TTabSheet", NULL))
```
dhrdlicka, line 233:
```suggestion
        FindWindowExW(winbox, NULL, L"TToolBar", NULL) &&
        FindWindowExW(winbox, NULL, L"TListBox", NULL) &&
        FindWindowExW(winbox, NULL, L"TStatusBar", NULL) &&
        (winbox = FindWindowExW(winbox, NULL, L"TPageControl", NULL)) && /* holds a TTabSheet even on VM pages */
        FindWindowExW(winbox, NULL, L"TTabSheet", NULL))
```

dhrdlicka (changes requested):

PR #4289, opened 2024-03-21
Modify the PCAP filter to allow multicast packets

cold-brewed:
Summary
=======
This PR modifies the PCAP filter accordingly:
* Multicast packets are now passed through
* The `broadcast` keyword is used instead of `ff:ff:ff:ff:ff:ff`

The biggest effect is fixing NetBEUI / NetBIOS which was not working before: NetBEUI uses a multicast address (`03:00:00:00:00:01`) instead of a standard ethernet broadcast which was not passing the PCAP filter.

Anything that uses NetBEUI / NetBIOS should now work with PCAP, including LAN Manager. Any other L3 that uses multicast should also now work.

Screenshot of win98 with no TCP/IP installed, using only NetBEUI to communicate with another win98 host configured the same:

netbeui

Checklist
=========

References
==========
N/A


NJRoadfan:
This should also fix AppleTalk if someone is using NT Services For Macintosh or PCMacLAN in a VM.

Issue #3577, opened 2023-08-15
Machine addition request (please post your machine addition requests here)


OBattler:
I am duplicating the Discord server post here, so people can use this issue to post their Machine addition requests in order to have them all in one place.

Write most machines requested on GitHub requested machines up here so I can pin this and close the issues:
- [saipan7] Acer AcerMate 600 (#3576);
- [saipan7] Dell OptiPlex GN+ (#3565);
- [pankoza2-pl] Packard Bell Legend 300sx (#3503);
- [saipan7] AST Advantage! Adventure 4066d (#3366);
- [pankoza2-pl] MSI MS-6199VA (retail and Packard Bell OEM) (#3338);
- [saipan7] Compaq Prolinea MT (#3325);
- [PechayClub] IBM PC Convertible (#3214);
- [TheDell1998] Shuttle HOT-597 (#3198);
- [TheDell1998] PB930/950, PB600, PB860, PB980, and PB992 (#3146, #3142);
- [MatthewH12] Sony Vaio / ASUS P2B-AE (#2206);
- [Banjo-Oz] Amstrad PC5286, PC7486SLC33, PC7486SLC80, and PC9486 (#2179);
- [swordlink1] Comapq Portable 486 (#2135);
- [headsetguy37] HP Falcon / ASUS MEB-VM (#1976);
- [Fenix770] Amstrad 5x86 PC machines (#1360);
- [CRTRevivalDreamer] TMC PCI54ST (#643);
- [jvernet] More 8086 systems (#505);
- [Spagheta-DSi] HP Vectra and HIL support (#377);
- [Mako88] EasyData 286 (Suntac 286-B) (#334).

lemondrops:
You might want to turn the list into a checklist (add `[ ] ` after each bullet), for easier tracking of what is done.

Tomeister1:
Hi do you know when all of these rom files will be able to use on 86box? If not can you please let me know when you can get to this task?

Thanks

TheDell1998:
I can't wait for the new machines to be here on 86Box!

Steveneska:
Is it too late to request another one? Not really a machine per se but a motherboard:

[...]

PR #4686, opened 2024-08-04
Add AOpen AX6LC machine

pcurtis1:
Summary
=======
Add AOpen AX6LC machine definition

Checklist
=========
* [x] Closes this comment for new machine: https://github.com/86Box/86Box/issues/3577#issuecomment-2267335467
* [x] ~~I have discussed this with core contributors already~~ I don't believe this is necessary in this case, but happy to discuss.
  * [x] I have opened a roms pull request - https://github.com/86Box/roms/pull/268

References
==========
Adapted from AX6BC which is very similar.

### DIFF
--- a/src/include/86box/machine.h
+++ b/src/include/86box/machine.h
@@ -803,6 +803,7 @@ extern int machine_at_lgibmx7g_init(const machine_t *);
 extern int machine_at_p3bf_init(const machine_t *);
 extern int machine_at_bf6_init(const machine_t *);
 extern int machine_at_ax6bc_init(const machine_t *);
+extern int machine_at_ax6lc_init(const machine_t *);
 extern int machine_at_atc6310bxii_init(const machine_t *);
 extern int machine_at_686bx_init(const machine_t *);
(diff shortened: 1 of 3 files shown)


pcurtis1:
@OBattler is there anything I can/need to do to help here?

OBattler:
What of unique does this machine bring? Seems to me like another generic Award BIOS machine.

OBattler:
This can be re-opened if I'm given a satisfying explanation.

pcurtis1:
> What of unique does this machine bring? Seems to me like another generic Award BIOS machine.
> This can be re-opened if I'm given a satisfying explanation.

The reason it's of value to me is because I own this board in real life - I don't own the other boards. It's valuable to me to be able to emulate the same hardware I own in real life. Having this in the main repo instead of having to maintain my own fork would be appreciated.

maxdlpee:
So, a user is offering to expand your machine selection and you reject that just because it is too similar to many other machines that are already in your project?
What sort of sorry excuse is that? We already got too many similar machines, and I don't see anyone complaining.

PR #4728, opened 2024-08-18
Add new machine GRiD GRiDcase 1520

techomancer:
AT/286 based early (1988) laptop with indestructible magnesium case.

Also add emulation of 3 specific Conner HDDs as required by stock 1520 BIOS. It only works with these particular drives and requires them to be somewhat slow or reads will hang. The alternative is to use RomBuster to patch your BIOS.

The machine had 3 display options (640x480 plasma or 2 different LCDs). For now we need to use stock CGA with mono amber monitor to get some of the plasma feel. Next step is to add Yamaha V6366 video card support.

Summary
=======
new machine as described

Checklist
=========
* [ ] Closes #xxx
* [ X] This pull request requires changes to the ROM set
  * [ X] I have opened a roms pull request - https://github.com/86Box/roms/pull/270/

References
==========

### DIFF
--- a/src/disk/hdc_ide.c
+++ b/src/disk/hdc_ide.c
@@ -558,7 +558,10 @@ ide_hd_identify(const ide_t *ide)
     /* Firmware */
     ide_padstr((char *) (ide->buffer + 23), EMU_VERSION_EX, 8);
     /* Model */
-    ide_padstr((char *) (ide->buffer + 27), device_identify, 40);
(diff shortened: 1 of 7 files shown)


techomancer:
Fixed all the errors/warnings. Removed dead code that remained from redundant high mem implementation (which is not necessary since 86Box has perfectly good default high mem implementation).

PR #4786, opened 2024-09-02
Use F8+F12 to release mouse on all platforms

chungy:
Summary
=======
It seems that Ctrl+End was used on non-Win32 platforms because WxWidgets couldn’t use non-modifier keys as a keyboard sequence.  Qt can, and F8+F12 interferes with the operation of software a lot less than Ctrl+End does (try doing some text editing!).

86Box hasn’t used WxWidgets for quite some time and this platform limitation hack has long outlived its necessity.

Checklist
=========
* [ ] Closes #xxx

### DIFF
--- a/src/device/keyboard.c
+++ b/src/device/keyboard.c
@@ -31,23 +31,13 @@
 
 int keyboard_scan;
 
-#ifdef _WIN32
-/* Windows: F8+F12 */
+/* F8+F12 */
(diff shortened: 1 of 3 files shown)


Digitoxin1:
Why not allow this to be reconfigured in the settings so if the default conflicts with something someone is doing with 86Box, they can remap it to something else.

Aerocatia:
Can this be merged?

chungy:
Are there any blockers on this?

PR #4973, opened 2024-11-17
Embed Qt base translations in the executable

OBattler:
Fixes standard messagebox buttons not being translated

Summary
=======

Checklist
=========
* [ ] Closes #xxx

References
==========


lemondrops:
Since the PR was created *a little* prematurely, here's the original description:

> Fixes standard message box buttons not being translated.

> Yes, this is technically a hack, but it's consistent with what we do for translations for 86Box itself, while allowing the executable to not depend on external translation files (while they can be bundled on macOS, on Windows they may not be present on runtime and bundling them is undesirable anyway).

PR #5189, opened 2025-01-30
Make HDD models' speeds accurate

MaxwellS04:
Summary
=======
Make HDD models (mostly from post-ATA-3 models)' random seek speeds accurate. Also change some, but few, of HDD models; names more accurate to its realistic versions. No HDD models have been added.

I heard that random seek speed for these HDD models which recently have been added are inaccurate, so I have opened a pull request for this.

Checklist
=========
* [ ] Closes #xxx

References
==========


unreal9010:
Great! Now we need some kind of UI modification so that the drives would correspond to their exact geometry and size.
At the moment, it's still possible to create these drives with customized geometry and sizes.

MaxwellS04:
> Great! Now we need some kind of UI modification so that the drives would correspond to their exact geometry and size. At the moment, it's still possible to create these drives with customized geometry and sizes.

Good thinking.

PR #5211, opened 2025-02-08
Add BEK-Tronic BEK-p407 machine type.

Enzoaperson:
Summary
=======
Adding the BEK-Tronic BEK-p407 486 motherboard as a select-able machine type. (Socket 3, SiS 496, Prime3B)

Checklist
=========
* [ ] Closes #xxx
  * [x] I have opened a roms pull request - https://github.com/86Box/roms/pull/302

References
==========
https://theretroweb.com/motherboard/manual/bek-p407-652aa211503a3636004898.pdf
https://theretroweb.com/motherboards/s/bek-tronic-bek-p407#chips


OBattler:
I don't really see the point of this - the only unique thing is the choice of Super I/O chip, otherwise, it's pretty much a standard Award SiS 496/497 machine and I think we already have enough of those, sorry. But I may add it in 5.0 or when it is that we expand the BIOS selection capability more widely.

PR #5227, opened 2025-02-13
Bring back the Slot 1 to Socket 8 adapter, but block it from use on incompatible machines.

pankoza2-pl:
Summary
=======
Instead of removing support of the Socket 8 to Slot 1 adapter completely, which was useful for running OSes that require ACPI or for using AGP Graphics Cards on Pentium Pro, just block the use of it on incompatible machines.

Checklist
=========


pankoza2-pl:
Note: blocking PII overdrive is not possible without blocking normal PII Deschutes due to their CPU type being merged, however it posts on VPC2007 still, but keep in mind hardware monitor error on the Asus P3V4X
There should be more flexibility in the CPU filter

PR #5346, opened 2025-03-14
Readd XT and AT floppy controllers

MaxwellS04:
Summary
=======
The title says at all.

Checklist
=========
* [ ] Closes #xxx

References
==========
They are removed for a year by a [commit](https://github.com/86Box/86Box/commit/abb066f6ef2efa9c83daad8ec50bcfb1fbcc29d8).

### DIFF
--- a/src/floppy/fdc.c
+++ b/src/floppy/fdc.c
@@ -105,6 +105,8 @@ static fdc_cards_t fdc_cards[] = {
     // clang-format off
     { &device_none               },
     { &device_internal           },
+    { &fdc_xt_device             },
+    { &fdc_at_device             },
     { &fdc_b215_device           },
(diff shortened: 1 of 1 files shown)


jriwanek:
There should be zero reason this is needed right now, considering machines which use them provide via device_internal (Choose Internal on the options) or via a SIO. can you state an actual reason for this PR?

MaxwellS04:
> There should be zero reason this is needed right now, considering machines which use them provide via device_internal (Choose Internal on the options) or via a SIO. can you state an actual reason for this PR?

Well, some of motherboards (mostly from the 8088 era to the Socket 1 era) even don't have internal floppy controller, according to The Retro Web, which a external floppy controller (mostly from some of Multi I/O controllers) is now needed to function.

PR #5367, opened 2025-03-19
Add JEGA (Japanese EGA) video card

akmed772:
Summary
=======
It adds an unpopular Japanese EGA (JEGA) video card emulation.

I wanted to make it generic, but the only available ROM is dumped from a laptop computer, Oki if386AX. The card initialization depends on the CMOS data of that machine, so I had to patch some part of the ROM. The original 16x16 font was not available from the contributor. I added [Shinonome 16-dot font](http://izumilib.web.fc2.com/sinonome/) that is available in free, but it doesn't have NEC's special characters. These ROMs might be replaced in the future.

I tested it with the Shuttle 386SX machine.

To contributors, thank you for providing ROMs and testing.

Checklist
=========
* [ ] Closes #xxx

References
==========
* https://en.wikipedia.org/wiki/AX_architecture
* Chips and Technologies 82C435 datasheet
* JEGA internal registers and I/O ports https://radioc.web.fc2.com/column/ax/jega_regs.htm

It seems there was an [official English version of AX Technical Reference Guide](https://sts.kahaku.go.jp/sts/detail.php?no=906390770026), but it's not available in public.

Screenshots
==========
Screenshot SE3
Windows 2.1


OBattler:
Merged. Question - can you upload the program from the first screenshot and the Windows 2.x for AX somewhere?

akmed772:
The first one is a free text editor for MS-DOS.
SE3 for AX color:
https://www.vector.co.jp/soft/dos/writing/se004934.html

Win2.1:
https://we.tl/t-dFam6ekeMR
pass:windows286

Issue #1639, opened 2021-08-30
Set Hotkey to release capture


wbsdty331:
**Is your feature request related to a problem? Please describe.**
It's a feature request.

**Describe the solution you'd like**
Custom HotKey to release capture, not just F8 + F12.
I always use VMware to use VM, and They use Ctrl + Alt HotKey to release capture and it can be modified in settings.
Maybe We can make a gui?

**Describe alternatives you've considered**

**Additional context**

hyanggi:
I think "F8 + F12" is not a good choice of the release hotkey. When using Windows 3.1, "F8" means "copy". Every time I press "F8" it opens the "Copy" dialog. It's annoying.

I think "Ctrl + Alt" is a more reasonable choice.

Try Windows 3.1 online: https://www.pcjs.org/software/pcx86/sys/windows/3.10/

Windows 3.0 has the same problem when using File Manager: https://www.pcjs.org/software/pcx86/sys/windows/3.00/

Nevon:
While perhaps niche, this becomes pretty important when using a virtual keyboard that may not have function keys or even if they do, it might not be possible to press both of them at the same time. My concrete example is the Steam deck in desktop mode.

brunocastello:
Letting the user choose a hotkey combo is probably the best available solution for every use case the 86box users from different platforms have. Current standard on macs opens the windows 9x start menu every time I attempt it.

OBattler:
The leading 1 means it's an E0-prefixed scan code.

Rast1234:
If anyone is looking to replicate VirtualBox host key behavior: release by right CTRL only. It will release only mouse but still good for muscle memory. 

`86Box.exe -P "path\to\vm" --keycodes 11D,000,000,000,11D,000`

PR #5482, opened 2025-04-18
Replaced hardcoded uncapture sequence with cfg directive and clarifie…

cathoderaydude:
Regarding #1639: added config directive escape_keys, which accepts either "f8_f12", "ctrl_end", or two hex scancodes in "0x000,0x000" format. Also updates status line above display if either preset is selected; otherwise says "Press custom key combo [...]"

While implementing, discovered existing --keycodes command line param with unclear usage. Replaced with new code that accepts 0x000,0x000 format and provides usage instructions.

Needs a GUI element IMO (I think a menu option in Action would be ideal), I can try to provide that in a future PR. Doesn't solve the longterm need for a robust key binding system, but it should be a stopgap. I am not a heavily experienced C/++ dev so hopefully my code quality is adequate, I tried to mirror the existing style.

Note: On my machine (windows 11) 86box does not appear to output stdin unless I manually "> stdin.txt", but as long as I do that the usage text does work.

### DIFF
--- a/src/86box.c
+++ b/src/86box.c
@@ -654,7 +654,7 @@ usage:
 #ifdef USE_INSTRUMENT
             printf("-J or --instrument name - set 'name' to be the profiling instrument\n");
 #endif
-            printf("-K or --keycodes codes  - set 'codes' to be the uncapture combination\n");
+            printf("-K or --keycodes codes  - set the uncapture combination (leave codes blank for instructions)\n");
             printf("-L or --logfile path    - set 'path' to be the logfile\n");
             printf("-M or --missing         - dump missing machines and video cards\n");
             printf("-N or --noconfirm       - do not ask for confirmation on quit\n");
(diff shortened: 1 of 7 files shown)


OBattler:
The lack of output is a Windows problem - to this day, for unknown reasons, Windows does not allow executables marked as graphics mode to output text to console. This is why I made my local compiles marked as character-mode so I can actually see the console output, unfortunately, it's that or having the program explicitly spawn a console, for which we do have an option, I believe.

cathoderaydude:
Oh, right, I forgot about that problem, ugh. Hmm. Specifically in the case of the usage instructions, it might make sense to sprintf the output into a buffer and then either printf the buffer if not on windows, or display it in a dialog if on windows. But that's a matter for another PR I suppose.

cathoderaydude:
Had a brainwave about how to reflect this setting in the UI without a big lift. There's a submenu under Actions now from which you can pick either of the preset options. If you set either preset in the config or command line it will be selected on launch. If you select a combo other than the two presets, the menu item will be greyed out and state that it's been customized.

This is incomplete however; it needs i18n support for one thing, but also, I couldn't figure out how to update the status line in the UI, so it doesn't reflect the current release combo if you change it after startup. Maybe this is an easy fix, otherwise maybe just consider the original commit.

OBattler:
I guess this can be closed as it is superseded by the PR I just merged.

PR #5489, opened 2025-04-20
Shortcut & keybind overhaul

### LINKED ISSUE #1639: Set Hotkey to release capture

cathoderaydude:
Summary
=======
This is a nearly complete user-customizable keybind system which obsoletes PR #5482 and resolves #1639. It adds:

* A global keybind table, acc_keys
* A lookup function, FindAccelerator(char name)
* A new keybind interface in the Input Devices configuration panel
* Config file save/load routines
* A framework on top of which an unlimited number of keyboard shortcuts can be added to the app.

This _removes_ the (undocumented) keycodes CLI param since it's superseded by the new system.

Issues
=======
* No i18n is in place because I don't know how any of that works.
* UI **IS** updated for menu shortcuts, but **NOT** for dialog strings or statusbar text.
* I have no idea if this works on Linux or Mac. Certainly the defaults will not work, since they use keys that don't exist. I think it should be simple enough to write some ifdefs to select a different default set of keys based on platform, you'd just need to find out what those defaults should be.

These are hairy issues and I suspect others would be able to tackle them much more quickly than I could. If nobody else has the time however, just let me know if there is interest in committing this contribution to the project. I can start investigating them myself, I have Linux and Mac systems available and would be happy to look into it, I just want to make sure I'm spending my time wisely.

I should point out that this code depends on Qt KeySequence, without which it would have been massively greater effort to implement. In my opinion however, KeySequence is a terrible, clunky, limited system. In particular, there is simply _no_ practical way to convert a KeySequence to a set of scancodes short of writing your own (extremely tedious) conversion routine, so it will not be possible to use this code from any non-Qt-aware portions of the codebase.

I don't _think_ this matters - the only shortcut I'm aware of that was in the C portion was the mouse unbind key sequence routine, and to be frank I'm not sure why it was there, so I moved it over to mainwindow.cpp. And in general, since Qt runs the UI, this solution should work for just about any desired purpose.

Checklist
=========
* [ * ] Closes #1639

Screenshot 2025-04-19 200631

Screenshot 2025-04-19 200641


Dizzy611:
I can't speak for the team lead here but I can say among the core team we've been watching your earlier PR with definite interest and this new one is already a lot further than I expected you'd go with it.

jriwanek:
Definately not wasting your time, that's for sure.

Dizzy611:
Big cheese has given his okay as long as you can fix mac and linux building (currently failing GHA checks)

OBattler:
This actually looks awesome!

cathoderaydude:
> This actually looks awesome!

Thank you! Really glad to hear it.

Linux/mac build should be fixed; I had somehow missed dangling refs to the old escape key combo function. Also cleaned a couple other warnings that my compiler missed.

cathoderaydude:
I tested on Linux (Fedora 42, fresh install) and it appears to work perfectly with no changes. Testing on Mac (Ventura, opencore'd onto a 2015 mbp) explodes more or less as expected: the accelerators in menus appear as unicode question mark blocks, and the bind list is actually empty.

Curiously, I had the latter happen on Linux as well... when I ran the app as root accidentally. Even more curious is how this could happen at all; even if _all_ the shortcuts fail to parse, that has nothing to do with generating the list; everything stored in the acc_keys structure is plain text and should not be touching any of the QKeySequence nonsense. I gotta sit a spell and think about what could be going wrong in both of these cases.

Notably, the Mac version also segfaults on normal exit, and does not appear to write a config file, though I'm not sure where it's supposed to on Mac.

cathoderaydude:
And moments later the lightbulb goes on: The two issues are related. The acc_keys structure is initially prefilled with its default values within load_binds(), which I assumed was guaranteed to execute - but I am guessing that if the config file is unreadable, the entire config load process aborts and load_binds() never gets called. So I must find out why the config isn't writable in both of these cases, then contemplate where else I could init the table so this can't occur.

Cacodemon345:
> And moments later the lightbulb goes on: The two issues are related. The acc_keys structure is initially prefilled with its default values within load_binds(), which I assumed was guaranteed to execute - but I am guessing (and have just confirmed) that if the config file is unreadable, the entire config load process aborts and load_binds() never gets called. So I must find out why the config isn't writable in both of these cases, then contemplate where else I could init the table so this can't occur.

You can default-init the acc_keys structure in `pc_init_modules()`.

OBattler:
Also, I would recommend merging the combinations to go to and from full screen into one.

cathoderaydude:
> You can default-init the acc_keys structure in pc_init_modules()

Oh yeah, looks good, ty!

> Also, I would recommend merging the combinations to go to and from full screen into one.

So, that one seemed a bit odd to me - I was wondering why they were two different shortcuts to begin with, since they actually call the same method. And I'm not sure if this code was written by anyone here, so maybe my guesswork is unnecessary, but what it LOOKS like to me is that the original author ran into this problem:

1. The "fullscreen" combo works by using an accelerator attached to the fullscreen menu option
2. The menu accelerator context is unavailable when the window is fullscreen, so the un-fullscreen shortcut has to be bound to a different context.
3. Qt will not allow binding the same key combo to two different shortcuts (I tested this, it throws an "ambiguous shortcut" error and disables both.)

I have no doubt this can be resolved, it just requires some digging to figure out what is going on with the contexts. I appreciate your input though since I wasn't sure if this behavior was _desired_ or not.

cathoderaydude:
Since there's already an eventFilter installed, I used that to detect the screenshot combo in fullscreen - and, having done that, I realized that probably none of the _other_ shortcuts work in fullscreen, and sure enough that's true. Fixed that; now you can e.g. hard reset from fullscreen. Made sure the confirmation modal doesn't cause any problems.

OBattler:
So is this ready for merge now?

cathoderaydude:
> So is this ready for merge now?

99% of the way - I just finished adding support for translation and fixing some remaining bugs. I only have one remaining issue, which is that I can't figure out how to trigger ProgSettings::reloadStrings() from SettingsInput::save() in order to update the UI status line when the mouse release command is changed. If anyone can tell me how to solve that, I can give you a mergeable commit.

cathoderaydude:
Never mind - just had to include the header. Works now. Updated translations. Should be good to go.

OBattler:
OK, waiting for the workflows to run, then I'm going to merge.

PR #5496, opened 2025-04-22
Add write indicators for all drives

Cacodemon345:
Summary
=======
This PR adds write indicators for all drives. RX/TX activity is now also separately indicated for network cards.

Checklist
=========

References
==========
None.


src/qt/qt_iconindicators.cpp
```diff
@@ -24,9 +26,13 @@ getIconWithIndicator(const QIcon &icon, const QSize &size, QIcon::Mode iconMode,
         return iconPixmap;
 
     auto painter = QPainter(&iconPixmap);
-    auto indicatorPixmap = getIndicatorIcon(indicator).pixmap(size);
+    auto indicatorPixmap = getIndicatorIcon(indicator == ReadWriteActive ? Active : indicator).pixmap(size);
```
dhrdlicka, line 29: Can't we turn this into a bitfield if we are combining indicators?

dhrdlicka (changes requested):

PR #5539, opened 2025-05-02
Global config proof of concept

dhrdlicka:
Summary
=======
This PR introduces a global `86box.cfg` (filename may be subject to change) which contains all the settings configurable via the Preferences dialog (scope may be subject to change)

Checklist
=========
* [ ] Closes #xxx

References
==========


cathoderaydude:
I've been cogitating on this same problem for a bit, and I see what you're driving at, but I think you're running into the same issue I did, which is that I can't really come up with any setting other than "Language" that feels _unequivocally_ global. For everything else I can see cases where I'd want it to be global, and cases where I wouldn't.

Several of the settings in this PR are things I personally wouldn't want to be global, and conversely, it's short many settings I _would_ want (e.g. keyboard shortcuts, network interface bindings...) I think in the long term, it would lead to a lot of back-and-forth from people arguing that certain settings should be one or the other.

In my mind, the ultimate solution is to make _everything_ global-able by extending the existing config read/save logic with an extra step:

- On config load: For each ini directive, first check the machine config; if it's not found there, then check the global config; and if _that_ isn't found, fall back to hardcoded defaults.
- On config save: For each setting, check if there's a value in the global config; if so, and if the current value set in preferences differs from _both that and the hardcoded default,_ then write a directive to the machine config file.

The user would then be able to make any setting global by manually copying it into the global.cfg.

There are problems with this - obviously there'd be no UI visibility, and when a user makes a change that shadows a global setting, it won't be obvious and they won't be able to restore it without deleting the line from the machine's conf. I can't think of any sensible way to solve that however, and I think most people who would be interested in this feature at all would have no trouble keeping up.

Codewise, it's a mild pain in the ass to implement without some refactoring, but I think it would be a more flexible and sustainable solution long term.

PR #5546, opened 2025-05-04
New dynamic recompiler register optimizations

Cacodemon345:
Summary
=======
New dynamic recompiler register optimizations

1. Register enums now implicitly start from 0, instead of explicitly.
2. Removed unused `acycs` enum.
3. Volatile registers are now at the end of the list.
3. Make `codegen_reg_mark_as_required` function only handle registers after IREG_EBX and before any volatile registers.

3-5% performance improvement observed on Linux for Wolf3D MAPEDIT SETUP.

Checklist
=========

References
==========
None.


src/codegen_new/codegen_backend_x86-64_ops.c
```diff
@@ -68,8 +68,10 @@ jmp(codeblock_t *block, uintptr_t func)
 void
 host_x86_ADD8_REG_IMM(codeblock_t *block, int dst_reg, uint8_t imm_data)
 {
+#if 0
```
sskras, line 71: Just a thought: why not use some specific defined name in all of these ifdefs ?

This would make it easier to switch between the enabled and disabled states.

Cacodemon345: Done.

PR #5647, opened 2025-06-01
Rename the Olivetti M24 and M240 to add the names of their AT&T counterparts

pankoza2-pl:
Summary
=======
Because the Olivetti M24 was very widely known under the AT&T PC 6300 name, while the M240 was known as AT&T PC 6300 WGS, I changed the machine table to reflect these alternative names due to the big popularity of the AT&T counterparts

Checklist
=========

References
==========

### DIFF
--- a/src/machine/machine_table.c
+++ b/src/machine/machine_table.c
@@ -2377,7 +2377,7 @@ const machine_t machines[] = {
         .net_device = NULL
     },
     {
-        .name = "[8086] Olivetti M21/24/24SP",
+        .name = "[8086] Olivetti M21/24/24SP/AT&T PC 6300",
         .internal_name = "m24",
         .type = MACHINE_TYPE_8086,
         .chipset = MACHINE_CHIPSET_PROPRIETARY,
@@ -2417,7 +2417,7 @@ const machine_t machines[] = {
     },
     /* Has Olivetti KBC firmware. */
     {
-        .name = "[8086] Olivetti M240",
+        .name = "[8086] Olivetti M240/AT&T PC 6300 WGS",
         .internal_name = "m240",
         .type = MACHINE_TYPE_8086,
         .chipset = MACHINE_CHIPSET_PROPRIETARY,


OBattler:
From what I know the AT&T counterparts have their own separate BIOS'es, so someone would need to find them.

pankoza2-pl:
According to PCjs, The AT&T 6300 ROM BIOS is apparently identical to the Olivetti M24 ROM BIOS.

jester-xbmc:
Hi there, Olivetti M240 owner here (and avid BIOS dumper) so far I have doubts that the AT&T PC 6300 WGS BIOS = M240 BIOS, for example, see attached screenshot a 6300 WGS, it list BIOS 2.07, which is not existing on the M240, at least it is not mentioned in in documents nor the versions I have collected so far.

In short, i'd be careful and indeed, provide actual BIOS files for the AT&T machines so one could compare them
430695438_995179385449084_6278538237339173188_n

pankoza2-pl:
> Hi there, Olivetti M240 owner here (and avid BIOS dumper) so far I have doubts that the AT&T PC 6300 WGS BIOS = M240 BIOS, for example, see attached screenshot a 6300 WGS, it list BIOS 2.07, which is not existing on the M240, at least it is not mentioned in in documents nor the versions I have collected so far.
> 
> In short, i'd be careful and indeed, provide actual BIOS files for the AT&T machines so one could compare them
> 430695438_995179385449084_6278538237339173188_n
> 

If you have dumped this version, can you share it, I might try to add it to 86Box then

OBattler:
This is a European M240 (with Italian DOS) with 2.04 BIOS: 
tumblr_e9316a6df1254052312a3eb1a7703e9c_e5a19652_540
.

OBattler:
OK, I have compared the AT&T 6300 BIOS in VARCem to their Olivetti M24 BIOS - both are the exact same Olivetti 1.43 BIOS. So I'd say proceed with the rename.

PR #5914, opened 2025-08-09
Update German GUI translation

lotharsm:
Summary
=======
Minor fixes to the German GUI translation. The translation for the generic term "Images" was incorrect, e.g. it always referred to cart images even for CD-ROM images.

Additionally, I applied some minor typography and grammar fixes.

Checklist
=========
* [ ] Closes #xxx

References
==========

### DIFF
--- a/src/qt/languages/de-DE.po
+++ b/src/qt/languages/de-DE.po
@@ -13,7 +13,7 @@ msgid "&Keyboard requires capture"
 msgstr "&Tastatur benötigt das Einfangen des Mauszeigers"
 
 msgid "&Right CTRL is left ALT"
-msgstr "&Die rechte Strg-Taste ist die Linke Alt-Taste"
+msgstr "&Die rechte Strg-Taste ist die linke Alt-Taste"
 
 msgid "&Hard Reset..."
 msgstr "&Kaltstart..."
@@ -58,7 +58,7 @@ msgid "Qt (&OpenGL)"
 msgstr "Qt (&OpenGL)"
 
 msgid "Open&GL (3.0 Core)"
-msgstr "Open&GL (3.0-Kern)"
+msgstr "Open&GL (3.0 Core)"
 
 msgid "&VNC"
 msgstr "&VNC"
@@ -250,7 +250,7 @@ msgid "E&ject"
 msgstr "A&uswerfen"
 
 msgid "&Image..."
-msgstr "&Cartridgeabbild..."
+msgstr "&Abbild..."
 
 msgid "E&xport to 86F..."
 msgstr "&In das 86F-Format e&xportieren..."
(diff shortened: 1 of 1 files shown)


src/qt/languages/de-DE.po
```diff
@@ -1,10 +1,17 @@
...
+"PO-Revision-Date: \n"
+"Last-Translator: Lothar Serra Mari <mail@serra.me>\n"
+"Language-Team: \n"
+"Language: de_DE\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "X-Language: de_DE\n"
 "X-Source-Language: en_US\n"
+"X-Generator: Poedit 3.6\n"
```
nelsonhef, line 14: Please, do not use POEdit to edit translations. It breaks the formatting when there is double new line escapes (\n\n) and some other problems. Please use a plain text editor or GitHub editor.

nelsonhef (changes requested):

src/qt/languages/de-DE.po
```diff
@@ -684,8 +691,17 @@ msgstr "Wechseldatenträgerabbilder"
 msgid "Image %1"
 msgstr "Abbild %1"
 
-msgid "86Box could not find any usable ROM images.\n\nPlease <a href=\"https://github.com/86Box/roms/releases/latest\">download</a> a ROM set and extract it into the \"roms\" directory."
-msgstr "86Box konnte keine nutzbaren ROM-Dateien finden.\n\nBitte besuchen Sie <a href=\"https://github.com/86Box/roms/releases/latest\">download</a>, laden ein ROM-Set herunter und extrahieren dies in das \"roms\"-Verzeichnis."
+msgid ""
```
nelsonhef, line 694: Broken by POEdit. All msgid and msgstr lines should be in one line.

nelsonhef (approved):
Looks good now, thanks @lotharsm

lotharsm:
@nelsonhef Thank you, I re-did the changes with VS Code. Can you recommend an editor that can properly work around the multiple line breaks so I can easily check for fuzzy or missing translations in the future?

nelsonhef:
@lotharsm I mainly use nano for edits, but I have the same problem as POEdit with _msgmerge_, to merge an updated POT into the individual PO files, incorrectly interpreting double line breaks. What I do is, before processing the file with _msgmerge_, change all occurrences of ```\n\n``` to some other non-problematic string like ```*0*0```, make the changes, and then change ```*0*0``` back to ```\n\n```. Maybe you can try if this trick works with POEdit too.

When adding new translations, you can simply look for empties _msgstr ""_ and edit them accordingly.

lotharsm:
Thank you, replacing the "bad" characters is a really cool way, this should instantly solve the problems with PoEdit, and since we don't rely on the generated .mo files, it should work fine for editing purposes. I'll try this the next time.

PR #5935, opened 2025-08-11
Update Italian translation

Kappa971:
Summary
=======
I updated the Italian translation with the new strings.

I have some questions/considerations to make:
1. Is there a new virtual machine manager ?
It would be helpful to understand the context in which these new strings are displayed. For example, the meaning of "Display Name" is a bit ambiguous; in Italian it can be "nome dello schermo" (screen name) or "nome mostrato/visualizzato" (the name that is displayed).
2. Some new strings use the term "machine" or "VM," while others use "system". If they mean the same thing, wouldn't "system" be confusing?
3. I'm no audio expert, but the literal translation of "Nice Ramp" would be hilarious in any other language on the planet. 😂 
Sticking with Latin languages, what would be a suitable translation?
After doing some research, I think I've figured out that it refers to the amplification of some of the MT-32's instruments. Ramp perhaps refers to the waveform or something else, I don't know. It was previously translated as "Bella rampa", which seems like anything but a technical term 😆, so I changed it to "Amplificazione rampa migliorato" (Improved ramp amplification). 
I don't know if it's translated this way in Italian, or if it makes any logical sense, but I can't think of anything else...

### DIFF
--- a/src/qt/languages/it-IT.po
+++ b/src/qt/languages/it-IT.po
@@ -28,7 +28,7 @@ msgid "&Pause"
 msgstr "&Pausa"
 
 msgid "Pause"
-msgstr ""
+msgstr "Pausa"
 
 msgid "Re&sume"
 msgstr "&Riprendi"
@@ -130,7 +130,7 @@ msgid "&4:3"
 msgstr "&4:3"
 
 msgid "&Square pixels (Keep ratio)"
-msgstr "&Pixel quadrati (mantiene l'aspetto)"
+msgstr "&Pixel quadrati (mantieni proporzioni)"
 
 msgid "&Integer scale"
 msgstr "&Scala intera"
@@ -196,7 +196,7 @@ msgid "&Settings..."
 msgstr "&Impostazioni..."
 
 msgid "Settings..."
-msgstr ""
+msgstr "Impostazioni..."
 
 msgid "&Update status bar icons"
 msgstr "&Aggiorna icone della barra di stato"
@@ -364,7 +364,7 @@ msgid "Configure"
 msgstr "Configura"
 
 msgid "CPU:"
-msgstr ""
+msgstr "CPU"
 
 msgid "CPU type:"
 msgstr "Tipo di CPU:"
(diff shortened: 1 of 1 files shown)


Dizzy611:
1. Yes. It's in beta, and can be used by running 86Box with a command line argument of `-E folderhere` where "folderhere" is the path to a folder in which you keep your VMs. The correct translation would be the latter (name that is displayed) as this is the friendly name that is displayed for a VM as opposed to the name of the folder for said VM.
2. Can't really answer this one as I don't know why those three different terms are used and don't make those kinds of decisions.
3. I think you mostly made the right call here. The "Ramp" in question afaik refers to an effect where the volume of sound is gradually increased or decreased. So, not knowing italian and thus not knowing how to correct your translation, but just using what you said it means, it would be more "Improved amplification ramp" instead of "Improved ramp amplification". That is to say, it is the ramp, which determines the amplification, that is improved, not the amplification itself.

nelsonhef:
@Kappa971 the DOSBox-X page explains this option as a smothing for sudden volume changes of instruments:
```
Toggles "Nice Amp Ramp" mode that improves amplitude ramp for sustaining instruments. Quick changes of volume or
expression on a MIDI channel may result in amp jumps on real hardware. When "Nice Amp Ramp" mode is enabled, amp
changes gradually instead. Otherwise, the emulation accuracy is preserved.
```

In the pt-BR translation, the best term that i found to this item is "Rampa Suave", which in Italian as per Google Translator is "Rampa liscia"

src/qt/languages/it-IT.po
```diff
@@ -1069,265 +1069,265 @@ msgid "Hard reset"
...
 
 msgid "Build"
-msgstr ""
+msgstr "Build"
 
 msgid "Version"
-msgstr ""
+msgstr "Versione"
 
 msgid "An update to 86Box is available!"
```
nelsonhef, line 1281: This msgid has been changed with PR #5933. Please change to msgid "\<b\>An update to 86Box is available!\</b\>" and msgstr "\<b\>È disponibile un aggiornamento per 86Box!\</b\>"

src/qt/languages/it-IT.po
```diff
@@ -364,7 +364,7 @@ msgid "Configure"
 msgstr "Configura"
 
 msgid "CPU:"
-msgstr ""
+msgstr "CPU"
```
nelsonhef, line 367: Missed a colon, should be "CPU:"

nelsonhef (changes requested):

Kappa971:
> 1. Yes. It's in beta, and can be used by running 86Box with a command line argument of `-E folderhere` where "folderhere" is the path to a folder in which you keep your VMs. The correct translation would be the latter (name that is displayed) as this is the friendly name that is displayed for a VM as opposed to the name of the folder for said VM.

It's a shame this VM manager is so hidden, I didn't know it existed. It should be the default startup for 86Box.

> 3. I think you mostly made the right call here. The "Ramp" in question afaik refers to an effect where the volume of sound is gradually increased or decreased. So, not knowing italian and thus not knowing how to correct your translation, but just using what you said it means, it would be more "Improved amplification ramp" instead of "Improved ramp amplification". That is to say, it is the ramp, which determines the amplification, that is improved, not the amplification itself.

"Improved amplification ramp" in Italian would be "Rampa di amplificazione migliorata". It's a bit long, is the MT-32 configuration window fixed size or does it resize based on the length of the text? I could also shorten it to "Rampa di ampl. migliorata".

> In the pt-BR translation, the best term that i found to this item is "Rampa Suave", which in Italian as per Google Translator is "Rampa liscia"

In Italian there is "soave" which is a more poetic way of saying "dolce" (sweet) or "delicato" (delicate), but it sounds bad in this context. "Rampa di amplificazione graduale" might sound better.

src/qt/languages/it-IT.po
Kappa971: Done

src/qt/languages/it-IT.po
Kappa971: Done

Kappa971:
I downloaded the latest build and the new translated strings are not there.

nelsonhef:
@Kappa971 run with the following parameters to show the VM Manager in Italian: _86box,exe -E \<vmfolder\> -G it-IT_

Kappa971:
Issue fixed with build 7480.

PR #5973, opened 2025-08-16
Update Italian translation

Kappa971:
Summary
=======
Small fixes.

### DIFF
--- a/src/qt/languages/it-IT.po
+++ b/src/qt/languages/it-IT.po
@@ -478,7 +478,7 @@ msgid "MIDI Out:"
 msgstr "Uscita MIDI:"
 
 msgid "Standalone MPU-401"
-msgstr "MPU-401 autonomo"
+msgstr "Scheda dedicata MPU-401"
 
 msgid "Use FLOAT32 sound"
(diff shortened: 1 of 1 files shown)


src/qt/languages/it-IT.po
```diff
@@ -2449,7 +2449,7 @@ msgid "Composite"
 msgstr "Composito"
 
 msgid "True color"
-msgstr ""
+msgstr "Milioni di colori"
```
nelsonhef, line 2452: This [True Color](https://github.com/86Box/86Box/blob/c9d2b60dbe0b363df9cdde1d3552448281c4eb40/src/video/vid_cga_v6355.c#L608) string was added with the Yamaha V6355 CGA video chipset. It's not related with the number of colors (still the 16 color CGA) but with the accuracy of color.

nelsonhef (changes requested):

src/qt/languages/it-IT.po
OBattler: The Yamaha V6355 does in fact have a mode that allows mapping of the 16 colors to arbitrary colors.

PR #6009, opened 2025-08-22
Fixes for broken translation

usergithub64:
Summary
=======

Checklist
=========
* [ ] Closes #xxx

References
==========


lemondrops:
This would cause the strings to be doubly translated - a better fix would be to add a colon to the affected strings in the *.po and *.pot files.
<https://github.com/86Box/86Box/blob/master/src/qt/qt_vmmanager_detailsection.cpp#L180>

usergithub64:
Then, when compiling, a message appears about the duplicate of the line in the .po file. Since there is already "ISA RTC:" There is other peripherals in the tab.

lemondrops:
Let me handle this, then; I have a PR in the works.

PR #6101, opened 2025-08-29
use SO_REUSEADDR on gdb stub socket

sherpya:
Summary
=======
I'm trying to debug a real bios with gdb stub, I restart often and sometime 86Box doesn't wait for debugger.
Looks like the problem is missing SO_REUSEADDR on the socket.
Could you please check on other platform (it's only tested on Linux)

### DIFF
--- a/src/gdbstub.c
+++ b/src/gdbstub.c
@@ -1791,6 +1791,24 @@ gdbstub_init(void)
         return;
     }
 
+    int yes = 1;
+    if (setsockopt(gdbstub_socket, SOL_SOCKET, SO_REUSEADDR,
+#ifdef _WIN32
+                   (const char *) &yes,
+#else
+                   &yes,
+#endif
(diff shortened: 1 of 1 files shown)


Cacodemon345:
You should ifdef out the part for Windows platforms; this should be fine otherwise.

sherpya:
> You should ifdef out the part for Windows platforms; this should be fine otherwise.

windows has SO_REUSEADDR I think the difference is in the function signature (it needs const char * instead)

sherpya:
I've updated the code path for windows (also added SO_EXCLUSIVEADDRUSE), but I'm unable to trigger the action for msys2 in my fork

OBattler:
I approved the actions on your pull request now, so the one for Msys2 is going to be performed here now. Waiting for them all to finish now.

OBattler:
It passed all checks, therefore, merged.

PR #6107, opened 2025-08-29
Fix Force 4:3 Aspect Ratio not working if Resizable window is enabled (#6062)

### LINKED ISSUE #6062: Minor: Force 4:3 Aspect Ratio not working if Resizable window is enabled
### Checklist

- [x] I **have searched** the issue tracker and **was unable** to find an [open](../issues?q=is%3Aissue+is%3Aopen) or [closed](../issues?q=is%3Aissue+is%3Aclosed) issue matching what I'm seeing.
- [x] I **have verified** that the issue is happening in the **[latest nightly build](https://ci.86box.net/job/86Box/lastSuccessfulBuild/artifact/)**, or the latest commit, if building from source.

### What happened?

When running the machine, in View menu, **Force 4:3 Aspect Ratio** not working if **Resizable window** is enabled. It could allow resizing respecting 4:3 or falling back to 4:3 after resizing.

Affects v5.0 and 5.0.1 
Any renderer.

### Configuration file

```ini
Any machine seems to show that behavior.
```

### Operating system

Windows 11

### 86Box version

5.0.1 7624

### Build architecture

Windows - x64 (64-bit)

### Build type

- [ ] New recompiler
- [ ] Debug build

### Additional context

_No response_

andresdelcampo:
Summary
=======
I finally decided to give this a go myself: [… #6062](https://github.com/86Box/86Box/issues/6062)
Summary of the issue is that force aspect 4:3 is ignored when resizing window is enabled. 

AI assisted fix. Its explanation (seems reasonable): 
_The root of the bug is that 86Box’s UI only applies the 4:3 constraint when it handles its own resize requests. In qt_mainwindow.cpp the function MainWindow::resizeEvent(QResizeEvent *event) returns immediately whenever the window is either in fullscreen or the Resizable window mode (vid_resize==1). When this early return is hit the program does not emit a resizeContents signal, so the code that adjusts the stack‑widget size and calls setFixedSize in the constructor’s lambda is skipped. The only place where the 4:3 option force_43 is used in the Qt layer is to tick the menu entry and toggle the flag via video_toggle_option; neither resizeEvent nor the resizeContents handler look at force_43. Consequently, when Resizable window is enabled the user can drag the window to arbitrary aspect ratios, and the forced 4:3 mode is silently ignored._
_A practical fix is to enforce the aspect ratio in the UI whenever the user resizes the window while force_43 is active._

Tested in Windows 11 WSL Ubuntu compiling Linux -a bit shaky to resize windows in this environment by default, but seems to behave fine now -it needed a 2nd iteration to get the fix right. Verified checking/unchecking Force 4:3 aspect and Resizing window.
Also verified two machines opened at the same time. 
I would recommend additional testing if the fix is accepted as I am not versed in Qt and could only test one mixed config. I can help further re-testing Windows -preferably getting a built executable.
Either way, I am responsible for the fix and would follow up within a day max if something needs to be changed or the fix reverted.

Checklist
=========
* [X] Closes #6062
* [NO] I have discussed this with core contributors already
* [N/A] This pull request requires changes to the ROM set
* [N/A] I have opened a roms pull request - https://github.com/86Box/roms/pull/changeme/

References
==========
N/A

### DIFF
--- a/src/qt/qt_mainwindow.cpp
+++ b/src/qt/qt_mainwindow.cpp
@@ -171,6 +171,8 @@ extern "C" void qt_blit(int x, int y, int w, int h, int monitor_index);
 
 extern MainWindow *main_window;
 
+bool MainWindow::s_adjustingForce43 = false;
+
 MainWindow::MainWindow(QWidget *parent)
     : QMainWindow(parent)
     , ui(new Ui::MainWindow)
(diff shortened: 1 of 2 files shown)


OBattler:
This doesn't work correctly, at all, it resizes the window to 4:3 instead of resizing the contents, and on the next resize, it's no longer forced to 4:3.

andresdelcampo:
Tried a few more iterations but this time in MinGW Windows. 
This one works fine in my setup as proven by https://1drv.ms/f/c/2acf3a53da6a23a0/EiOMC9wZtZlGmNYQeEg5iroBpA9HssTthbYys79PChAYrw?e=PDQYJb
but I am not so sure what you saw in the first iteration or in which system.

OBattler:
I'm on Windows 11. But anyway, waiting for the new version to compile so I can test it.

andresdelcampo:
Thanks either way for your timely responses and for following up on this. My compromise to own the fix stays -if something goes wrong, I will take the hit to fix or undo anything needed fast.

OBattler:
OK, now it works!

PR #6125, opened 2025-09-03
Added more controllers.

Gorilli09:
I've added more controller that have been pre-installed on Windows 95, 98, and 2000. I've also added an experimental controller called the "Ultimate Controller". It maxes out all of the inputs and is only used for testing purposes. Sorry, I'm not a real programmer. The reason why I edited this because I want to get an idea.


jriwanek:
I'm sorry to say but this code is completely useless, functions arn't implemented, code is incorrect, does nothing of use, and for the few which are even vaguely correct, I already have implementations on my branches.

Gorilli09:
WHY DO EVERY TIME I MAKE PULL REQUESTS, EVERYONE CLOSES THEM LIKE I'M BEING HATED!?

marie0x00:
lol

OBattler:
You were literally given an explanation above as for why this one was closed.

jriwanek:
When you make a PR which actually does something useful, then we'll be happy to merge it, until then, no, we're NOT going to merge broken code which does nothing of use.

PR #6166, opened 2025-09-14
Floppy disks sound emulation prototype for reviewing

Domppari:
Summary
=======
Floppy disk sound emulation prototype for review. Emulates 3.5" 300RPM 1,44MB FDD and 5.25 300 RPM 1.2MB spindle motor sound, spin-up, spin-down and spinning. Emulates single seek sounds during disk read/write and multitrack seek. Supports concurrent sound emulation for all active FDDs. 

Fixes issue with FDC where, when seeking, the drive status was immediately released from "busy" state, which caused BIOS to fire next commands for the drive, even when the seek was not done. The fix can be observed in POST FDD test, as it will take longer for testing drives than before (if enabled in BIOS). This is a fix for FDC emulation and not purely related to the FDD sound emulation. This fix isn't disabled by the -DDISABLE_FDD_AUDIO flag (mentioned later).

Added checkbox to enable Floppy disk drive sounds for all drives in floppy disk settings, and loading and saving for the selection. This is a shared setting for all drives and is not drive specific.

Feature is on by default in the cmake build, but can be disabled with flag -DDISABLE_FDD_AUDIO, which will disable the feature code and hide the selection from UI settings.

Notes:
- Some simple audio mixing is done in the fdd_audio.c but the basic function is to add audio samples for playback and then similar audio thread that cd audio uses will callback the sampled data. Should not be too taxing for the CPU, as it is mostly just data transferring for the audio buffers.
- Emulation assumes currently maximum track seek length to be 80 tracks, this is incorrect and over 80 track seeks only plays 80 tracks of seek sound sample.
- Sound emulation samples are 48kHz, 16bit mono PCM WAVE-samples and will be send to audio system without any resampling to different sample rate. If the audio system uses different playback sampling rate, this should be fixed to do resampling, or edit separate samples for different sample rates.
- Tested only on mingw64 Windows build. 
- Sound emulation currently detects only 3.5" and 5.25" drives for emulating,.

Known Issues:
- If floppy disk sound emulation setting is changed when VM is running, restart does not recognize the setting changes. The VM needs to be shutdown and cold started.

Checklist
=========
* [x] Pull request comes with required audio samples currently in the same directory with the 86box executable

References
==========
Used sound samples have been editor from http://shiru.untergrund.net/files/flopster.zip
Original samples can edited and used separately from the Flopster VSTi under the
Creative Commons Attribution (CC-BY) license terms.


OBattler:
> _Emulation assumes currently maximum track seek length to be 80 tracks, this is incorrect and over 80 track seeks only plays 80 tracks of seek sound sample._

That's corroborated by the FDC datasheets - any seek beyond 80 tracks needs a second seek.

Also, you added a seek complete interrupt but that's wrong - seeks do *NOT* interrupts, that's why the SENSE INTERRUPT STATUS command exists.

Edit: I may be wrong about the 80 tracks (one byte would indicate up to 255 tracks, though the PC87306 supports an extended mode with up to 4095 tracks), though that's for seek, recalibrate *is* limited though, to 77 or 80, depending on the controller.

Domppari:
> > _Emulation assumes currently maximum track seek length to be 80 tracks, this is incorrect and over 80 track seeks only plays 80 tracks of seek sound sample._
> 
> That's corroborated by the FDC datasheets - any seek beyond 80 tracks needs a second seek.
> 
> Also, you added a seek complete interrupt but that's wrong - seeks do _NOT_ interrupts, that's why the SENSE INTERRUPT STATUS command exists.

Ok, good to know about the FDC handling for over 80 track seeks - as we won't be needing contiguous seek sample over 80 tracks.

I'll revisit the code that handles the seek complete part. So, basically the BIOS will read the SENSE INTERRUPT STATUS, when the seek is ready. So I would only need to lift the drive not busy flag and that would solve the issue. Hopefully.

OBattler:
Wait, I was wrong! Seek *does* interrupt:
```C
        case 0x0f: /*Seek*/
            fdc->st0  = 0x20 | (fdc->params[0] & 3);
            fdc->stat = 0x80 | (1 << fdc->rw_drive);
            if (fdc->flags & FDC_FLAG_PCJR) {
                fdc->fintr     = 1;
                fdc->interrupt = -4;
                timer_set_delay_u64(&fdc->timer, 1024 * TIMER_USEC);
            } else {
                fdc->interrupt = -3;
                fdc_callback(fdc);
            }
            return;
```

Nevermind then.

OBattler:
Also, I was wrong - the SEEK command does not appear to be limited to how many tracks it can seek, only recalibrate is (it's fixed to 77 or 80 tracks, depending on the controller).

OBattler:
So your changes are actually fine as is, just test them with the PCjr to make sure that doesn't break!

Domppari:
Ok, good! I'll check PCJr!

OBattler:
Oh, and also check the PS/1, PS/2, and PS/55 machines, since they can also be finicky about things.

Domppari:
It seems that actually it did break at least the PCJr. The FDC code for PCJr did not change, so the issue would be in fdd.c -code. I'll try to find what causes this.

OBattler:
It did change - your seek callback does not have the PCjr-specific behavior.

OBattler:
Furthermore, I think it interrupts twice - once in fdc_poll(), summoned with:
```C
timer_set_delay_u64(&fdc->timer, 1024 * 1024 * 10 * TIMER_USEC);
```
And once in the FDD seek callback. I think with the FDD seek callback, this set_delay is not needed at all.

Domppari:
Yep. It did interrupt twice. Removed the second for normal AT/etc.

For PCJr, if I just exit the new interrupt callback function without any action, the PCJr acts fine (as it did before). 
```
void
fdc_seek_complete_interrupt(fdc_t *fdc, int drive)
{
    if (!fdc) {
        fdc_log("ERROR: fdc_seek_complete_interrupt called with NULL fdc!\n");
        return;
    }

    fdc_log("FDD %c: Seek complete interrupt\n", 0x41 + drive);

    if (fdc->flags & FDC_FLAG_PCJR)
        return;
```

The only problem is that, it does not wait for seek to finish and BIOS fires away commands to it. So in some part of the implementation the PCJr FDC sets some flags or something that tells the BIOS that the seek is ready. Need to find that section and alter it a bit -> move that part to the fdc_seek_complete_interrupt.

Might be an interrupt that shouldn't be fired until the seek is finished.

OBattler:
I see this in the code:
```C
    // Count actual seek time (6ms per track + 50ms base)
    // 80 tracks -> 50 + 6 * 80 = 530ms
    uint64_t seek_time_us = (50000 + (abs(track_diff) * 6000)) * TIMER_USEC;
```
For a thick headed drive (360k, which the PCjr expects), seek times should maybe be twice that? So 12 ms per track, since the head covers twice the distance on each track. And if this is based on 3.5" floppies, then this needs to be scaled as well, by 5.25/3.5 = 1.5, since bigger medium means bigger physical track. So this would mean 9 ms per track for a 5.25" 720k (2QD) or 1.2M (2DD) drive, and 18 ms per track for a 5.25" 160k/180k (1DD) or 320k/360k (2DD) drive. 6 ms would then be for 3.5" 720k (2DD), 1.44M (2HD), or 2.88M (2ED) drive.

Domppari:
That might be part of, but currently it seems, that because FDD-code immediately sets the new track as current track, and PCJr BIOS seems to want to know on which track the head currently is - i think it's polling it. So when the target track is immediately set, it's very happy about it and continues to next commands.

I tried to set new drive->target_track field and when the seek-timer fires, just then the target track is set. This was not enough, it seems the PCJr does not like that the current track does not change within some time period and just bails out and loads the BASIC ROM.

So it seems that I would probably need to advance the current_track with timer to keep the PCJr BIOS happy.

That's my assumption on my tests today.

Kinda makes sense, as this is what a real drive would actually perform.

I'll test this hypothesis tomorrow.

OBattler:
If it was that, then it'd break on the current 86Box code as well, but it doesn't.

OBattler:
The current 86Box code, for PCjr, sets fdc->fintr to 1 (which allows Sense Interrupt Status to operate correctly), then delays with the timer, your code does not - it only sets fdc->fintr to 1 before the delay. I think that's breaking it.

Domppari:
If you mean this part, it's still there. The sis would be active and it would detect the track change almost instantly?

image

Domppari:
Ok, so the PCJr (tested with 5,25" 360k) seeking works now and normal PC's too (tested with 5,25" 360kt and 3,5" 1,44MB). The correct time period as waited until next seek occurs. Not really sure if those changes are in line with anything, but that change did do the trick.

Also the actual time for seek/drive is only an assumption. The 6ms come from the Mitsumi 1,44MB drive, but other drive seek periods might and would vary.

Have to check the rest of the systems too.

EDIT:
Now the PS/1 is broken... I think I need to compare to logs without and with the seek periods

Domppari:
Now also PS/1, PS/2 and PS/55 works correctly.

Domppari:
I'll do more system testing today for older XT/AT systems and multiple drives.

Domppari:
Following systems tested with older ones using 360k 5,25 drives and newer one using 1,44MB 3,5 drives. Also tested all the selectable floppy disk controllers (ISA-versions for older system) and compared to vanilla 86box 5.0 run.

All seem to act without a problem.

image

Domppari:
Added loading of samples from executablePath/samples and if does not find, then tries executablePath.

Domppari:
I need to retest all systems, as the timings changed.

Domppari:
PCJr required longer seek-times. All systems working correctly now.

Domppari:
Fixed the second FDD motor kept running, when only first drive was accessed.

PCJr still has odd timing issues. Usually it works fine, but sometimes at failes to load disk. I need to tweak the timings more for it.

OBattler:
The motor always runs when the relevant motor ON bit in the DOR register is enabled, at least from what I understand.

OBattler:
This is from the Winbond W83877TF datasheet:
imagem

OBattler:
And as I said, Microsoft DMFWRITE.EXE actually relies on that - it turns on both drives' motors at the same time and then switches between them.

Domppari:
I'm not sure if you understand the problem here correctly. Each drive's motor can run separately and individually and can be commanded to do so, as stated by the documentation provided. The issue here is that, when I access drive B: and it starts the motor, then I access the drive A: before the drive B: motor is commanded off, the A: motor starts. At this point both drive's motors are running. If I then keep accessing _ONLY_ the drive A: like for 1 minute, the drive A: motor keeps spinning for that minute, but so keeps drive B: motor. That is incorrect. This is what is fixed now. The problem could be observed in BIOS POST test routine, which caused incorrect spinning behavior in vanillla 86box implementation when using two drives. Also this could be observed in DOS, by accessing drive B: then back to drive a: and then just pressing enter on drive a: prompt would keep drive A: motor running, but it did also keep B: drive motor running.

Without audio cues, you wouldn't notice this at all, unless you'd look in the FDD/FDC logs very closely. It did not affect the technical behavior of the drive but was apparently wrong when sound was added.

Of course, could be that I'm missing the point here.

OBattler:
I just looed at the code changes and... the new DOR behavior seems to be basically the same as before, except for the indexes, and the motor not being disabled if it was already disabled before.

Domppari:
I think the issue was with drive swapping. We were using wrong index at some situations, which would leave some drive motor bit as not cleared.

This or that, based on my tests it functions as expected and the drive motor shutdowns work correctly when there is no accessing the drive for a second or two - this is controlled by the BIOS, the drive index was incorrectly used / determined.

Domppari:
> I just looed at the code changes and... the new DOR behavior seems to be basically the same as before, except for the indexes, and the motor not being disabled if it was already disabled before.

My bad. Don't know what happened before, but both drives still seem to spin during boot, even when B-drive is not accessed during that time.

OBattler:
I suspect that this behavior varies from BIOS to BIOS - some switch drive motors on and off when probing different drives, some have them all spinning at the same time.

jriwanek:
@Domppari So, for my notes, what's left to be done before this is ready for merger?

Domppari:
I'll add TeacFD 5.25" drive sound samples. Make UI changes to FDD settings, so that you can select what sounds do you want for each drive separately. The selection would include three drives: Generic Mitsumi 3.5" 1.44MB drive, Panasonic 5.25" 1.2MB drive and Teac 5.25" 1.2MB drive. Some small timing tweaking still do be done and additional system testing.

I'll tell you when I think it's ready for merge from my side.

Domppari:
Teac 5.25" drive samples added and implemented drive specific audio selection to FDD settings. Changed the audio sample path to look into roms/samples/fdd instead of '/samples' and root folder.

So for me, the PR is ready to merge at your will. But the samples should be moved to the roms repository and removed from 86box  source - if that's ok. Should be do it now, or?

jriwanek:
Better now than later imo

OBattler:
CodeQL found a problem here: https://github.com/86Box/86Box/pull/6166#discussion_r2366382997 .

OBattler:
I tried this but I can't hear the sounds - how do I enable them?

Domppari:
> I tried this but I can't hear the sounds - how do I enable them?

Copy samples to roms/samples/fdd folder and set audio for drive in the fdd settings.

PR #6233, opened 2025-09-27
Change window resizing logic when using 4:3 aspect ratio

andresdelcampo:
Summary
=======
With the current 4:3 resizing logic, maximizing the window would maximize horizontally and leave part of the window outside the screen. It seems to be desired that the resizing happens in the content -like PCEm- and not the window. See new comment in previous PR https://github.com/86Box/86Box/issues/6062#issuecomment-3336805338 requesting the change.

The previous approach could also become glitchy, especially when trying to reduce the window size, which is solved by this new approach.

Therefore, this PR changes window resizing logic when using 4:3 aspect ratio to **resize content only instead of resizing the window**. Also fixed issues in Remember size and position that are derived from the change. 

**Known issues:**
* Present in 86Box 5.0: When opening a VM with Resizable window and no Remember size, the window appears at 320:200 -stretched or not depending on 4:3 ratio -or always stretched in 5.0.
* Present in 86Box 5.0: When in fullscreen, always uses 4:3 ratio independently of the menu option. Ok with me... 
* There is a slight flicker while resizing with force 4:3 aspect ratio.

**Tested**
In Windows with multiple combinations of 
* Force aspect 4:3
* Remember size and position -and reopening the machine 
* Resizable screen - with and without force aspect and remember size, etc.
* Fullscreen -back and forth with different settings

Checklist
=========
* [X] Closes #6062 -thought it is closed already
* [X] I have discussed this with core contributors already - requested by @arrowgent
* [ ] This pull request requires changes to the ROM set
  * [ ] I have opened a roms pull request - https://github.com/86Box/roms/pull/changeme/

References
==========
_Provide links to datasheets or other documentation that helped you implement this pull request._

PR #6239, opened 2025-09-27
Fix remember size and position regression in which vertical size could grow when reopening the VM

andresdelcampo:
Summary
=======
Fixes the regression introduced in https://github.com/86Box/86Box/pull/6233 in which the vertical size could keep growing on VM reopen when remembering size and pos, due to not removing from the height the showed or hidden non-content elements (toolbar and status bar). 

This fix ensures they are taken into account for the vertical size. 

Tested with multiple combinations of parameters while reopening the machine, this time taking into account only the size of the content area. Of course including Hide Toolbar, Hide Status Bar and combinations of those while reopening the VM and ensuring the size remains.

Unfortunately it is extremely hard to ensure that this functionality works correctly in all cases as there are too many parameters that interact with each other sometimes even in contradicting ways. That said, I can just offer to jump in whenever anything is found related to any of my changes to fix it to the best of my ability.

Checklist
=========
* [ ] Closes #xxx
* [X] I have discussed this with core contributors already -reported by @OBattler in the former PR per Discord comment from gr

References
==========
(https://github.com/86Box/86Box/pull/6233)
https://discord.com/channels/262614059009048590/1421479562029830154


OBattler:
And there's another regression, in high DPI modes with high DPI scaling, this happens:
imagem

When turning DPI scaling off and on again, this happens:
imagem

andresdelcampo:
Weird, I could swear I tried that setting -in fact I had it marked. But I have a repro. Fix coming shortly. If you can think of any other settings that could be affected apart from the ones considered so far let me know -to ensure there are no more regressions still present from the original fix.

andresdelcampo:
I don't get this. I reproed this. I have been trying for hours to fix this trying all kinds of things in code. Even switched back to earlier versions and try to remake it cleaner. After failing and coming back completely clean to latest and rebuilding from scratch... everything works as intended and I cannot repro this... I even created a new machine in case I added something to my config files... and nothing. I cannot repro it anymore with latest version recompiled clean. This might take a bit longer as I am quite puzzled right now.

andresdelcampo:
No, I don't get it. Neither with the latest development build. @OBattler  -or someone else, could you please try to reproduce this with the latest development build? Maybe we have seen some strange artifact due to a code upgrade that somehow goes away with a clean rebuild?

PR #6372, opened 2025-10-21
fix Display Type does not correctly apply to Video #2

bozoscum:
Summary
=======
mon_cga_palette is reset to 0 in function video_prepare() causes the 'Display Type' always be 'Default'
this solution might not be correct but works

Checklist
=========
* [ ] Closes #xxx

References
==========


jriwanek:
I'll look into a more correct fix, this looks to be more a problem with the config file handler

PR #6405, opened 2025-10-26
Unix mode OSD

Valefungo:
Summary
=======
this PR addresses a rarely used use case: running 86box on the linux framebuffer without QT and without X

running it that way is already supported but when doing it 86box loses all menu and all abilities to mount floppies and CDs, it also becomes the owner of the entire pc with no way of quitting it or changing virtual console.

I've added an on screen display menu that allows doing everything the textual console (src/unix/unix.c) does;
mount/unmount floppy,cd,...
hard reset the machine
quit 86box
seeing the current perf %

key bindings:
ALTGR-G opens the osd, it should not conflict with existing key combos, I hope
while is open:
arrows up, down moves the cursor
enter does the action, in mount it enters a list of appropriate files (*.img, *.iso)
ESC goes back to main or closes the OSD

among other things I had to implement touchpad support to be able to use the emulated Win on my laptop without X

this was tested on a debian VM running on my main pc, on a Core-i7 8550U laptop and on a Raspberry Pi 4

attached is a photo of the OSD open running on the RPI

osd

Checklist
=========
* [ ] Closes #xxx

References
==========
nothing special

### DIFF
--- a/src/unix/unix.c
+++ b/src/unix/unix.c
@@ -1095,16 +1095,19 @@ unix_executeLine(char *line)
                         err = true;
                         fprintf(stderr, "Path name too long.\n");
                     }
-                    strcat(fn, xargv[curarg] + (xargv[curarg][0] == '\'' || xargv[curarg][0] == '"'));
-                    if (fn[strlen(fn) - 1] == '\''
-                        || fn[strlen(fn) - 1] == '"') {
-                        break;
+                    else
+                    {
+                        strcat(fn, xargv[curarg] + (xargv[curarg][0] == '\'' || xargv[curarg][0] == '"'));
+                        if (fn[strlen(fn) - 1] == '\''
+                            || fn[strlen(fn) - 1] == '"') {
+                            break;
+                        }
+                        strcat(fn, " ");
                     }
-                    strcat(fn, " ");
                 }
             } else {
                 if (strlen(xargv[2]) < PATH_MAX) {
-                    strcpy(fn, xargv[2]);
+                    strncpy(fn, xargv[2], PATH_MAX-1);
                 } else {
                     fprintf(stderr, "Path name too long.\n");
                 }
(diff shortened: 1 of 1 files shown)


Lili1228:
AltGr + anything is always a bad idea because that's how you type diacritics. On Latvian keyboard, AltGr + G (G is where W is on US keyboard) = Ģ, on Macedonian keyboard, AltGr + G = Г and so on.

Valefungo:
> AltGr + anything is always a bad idea because that's how you type diacritics. On Latvian keyboard, AltGr + G (G is where W is on US keyboard) = Ģ, on Macedonian keyboard, AltGr + G = Г and so on.

good point, didn't know at all.
suggestions for an alternative ?

dhrdlicka:
I suppose the best approach would be making this configurable via global config.

Cacodemon345:
Please fix build errors in the CI.

I suggest Alt+AltGr+G from a personal POV.

Valefungo:
> Please fix build errors in the CI.
> 
> I suggest Alt+AltGr+G from a personal POV.

Almost done.
I would go for Control-F11, it's the default hotkey for "make a screenshot" in QT mode
since in unix mode none of the QT hotkey combos are supported it currently does nothing.

is this ok for you ?

dhrdlicka:
Considering that the goal is an unescapable 86Box kiosk mode, this should ideally be some keystroke that's unlikely to be used by guest software. This pretty much leaves ordinary keystrokes like Alt+G or Ctrl+F11 out of the question.

Valefungo:
> Considering that the goal is an unescapable 86Box kiosk mode, this should ideally be some keystroke that's unlikely to be used by guest software. This pretty much leaves ordinary keystrokes like Alt+G or Ctrl+F11 out of the question.

found one nobody will never hit even by mistake:
Right Control - F11

OBattler:
It appears CodeQL has found several problems with the code.

PR #6419, opened 2025-10-29
Tandy: Implement CRTC-based display shifting (R2 & R7 registers)

pixel-jupiter:
Summary
=======
This PR adds support for shifting the display position using the CRTC's Horizontal Sync Position (R2) and Vertical Sync Position (R7). This enables the so-called the “screen shake” effect seen in many Sierra games, achieved by dynamically rewriting the specified registers on the fly.

[kq1_test.webm](https://github.com/user-attachments/assets/cd3478d8-3d16-4208-a66b-4039f8f45700)

A baseline calibration is used to distinguish between a game setting the initial screen position and a game intentionally shifting the screen for effect. After a video mode is set, the centered sync positions are stored as a reference. Every write to R2 or R7 triggers a calculation of the difference from the stored baseline.

Values outside the documented limits of these registers are not fully emulated:
- R2 accepts a wide range of values, with empty areas rendered as borders where active video is absent 
- R7 (just like in the original code) can produce visual glitches if it goes beyond its documented range (`R4 >= R7 >= R6`), but this only affects software that deliberately pushes the CRTC beyond spec to create special visual effects (for example - Area 5150)

Here's a test (within the documented range for R7) across three different modes: 80-column text, 320x200x4, and 320x200x16.

[basica_test.webm](https://github.com/user-attachments/assets/86a6f6fa-423f-4d25-999e-52901573b527)

All other modes available through BASICA have been tested as well.
640x200x16 for Tandy Video II (SL/2) was tested in a game that supports this mode.

Checklist
=========
* [ ] Closes #xxx

References
==========

### DIFF
--- a/src/video/vid_tandy.c
+++ b/src/video/vid_tandy.c
@@ -396,37 +396,51 @@ vid_render(tandy_t *dev, int line, int hos_offs)
     uint16_t  dat;
     int       col;
     int       cols[4];
+    uint32_t  border_val;
     int       out_x;
+    int       c_start;
 
     cols[0]       = (vid->array[2] & 0xf) + 16;
 
-    if (line >= 0) {
+    if (line < 0) {
+        if (dev->is_sl2 && (vid->array[5] & 1)) {
+            vid->memaddr += vid->crtc[1] * 2;
+        } else {
+            vid->memaddr += vid->crtc[1];
+        }
+        return;
+    }
+
+    if (vid->array[3] & 4) {
+        border_val = cols[0];
+    } else if ((vid->mode & 0x12) == 0x12) {
+        border_val = 0;
+    } else {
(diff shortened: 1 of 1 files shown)


OBattler:
I'm getting this warning now:
imagem
Where exactly do I put the parentheses?

pixel-jupiter:
Sorry about that - I didn't get any warnings when I built it on my end.

481
`vid->array[((dat >> (12 - (c >> 1) * 4)) & vid->array[1] & 0x0f) + 16] + 16;`

511
`vid->array[((dat >> (12 - (c >> 2) * 4)) & vid->array[1] & 0x0f) + 16] + 16;`

That should make it clear to the compiler.

OBattler:
Committed that and it does fix the warnings, thanks!

PR #6441, opened 2025-11-06
Rename the internal name of Goldstar CRD-8160B

Verloren50000:
Summary
=======
Rename the internal name of Goldstar CRD-8160B -- should be renamed from "goldstar" to "goldstar_8160b".

Checklist
=========

References
==========


OBattler:
No migration path provided so that would break existing configurations that use it.

PR #6709, opened 2026-01-13
MS-6117 corrections + onboard device overhaul

MaxwellS04:
Summary
=======
This PR corrects the FSB bus speeds and maximum RAM memory for the recently-added MSI MS-6117.

Also, this PR adds the missing "On-Board" names to the existing onboard video cards, `.vid_device` additions to remaining machines (`.snd_device` for Packard Bell PB8x0), correct the onboard sound on Gateway Tomahawk, and make the PS/ValuePoint 433DX/Si's onboard video configurable.

Checklist
=========
* [ ] Closes #xxx
* [ ] I have tested my changes locally and validated that the functionality works as intended

References
==========
[MSI MS-6117 User's Manual (for actual FSB bus speeds and maximum RAM memory)](https://theretroweb.com/motherboard/manual/ms-6117-user-manual-v1-0-63987fc3c68d0205900888.pdf)

### DIFF
--- a/src/video/vid_jega.c
+++ b/src/video/vid_jega.c
@@ -1042,7 +1042,7 @@ if386jega_available(void)
 }
 
 const device_t if386jega_device = {
-    .name          = "JEGA (if386AX)",
+    .name          = "JEGA On-Board (OKI if386AX30L)",
     .internal_name = "if386jega",
     .flags         = DEVICE_ISA,
     .local         = 0,


sentogahara:
The manual has an error and contradicts itself as it says the board has three RAM slots and supports up to 256 MB EDO/SDRAM DIMMs, meaning it should support up to 768 MB with all three slots populated, yet it also says it supports up to 1 GB EDO RAM just before that. Since, at least to my knowledge, the 440LX chipset (and also its successor, the 440BX) indeed only supports up to 256 MB DIMMs, I think the limit should be changed back to 768 MB.

sentogahara:
Also I don't think 384 MB EDO DIMMs (which would be the closest proper size as 1024 MB cannot be fully divided into three) even exist, further proving the manual is incorrect.

EDIT: Unless I missed something, it seems 384 MB SDRAM DIMMs don't exist at all, let alone EDO DIMMs. SDRAM DIMMs exist in 256 and 512 MB size, but not 384 or any other similar size like 320.

MaxwellS04:
> The manual has an error and contradicts itself as it says the board has three RAM slots and supports up to 256 MB EDO/SDRAM DIMMs, meaning it should support up to 768 MB with all three slots populated, yet it also says it supports up to 1 GB EDO RAM just before that. Since, at least to my knowledge, the 440LX chipset (and also its successor, the 440BX) indeed only supports up to 256 MB DIMMs, I think the limit should be changed back to 768 MB.

I understand your knowledge, but I believe the MS-6117's user manual thinks that its maximum memory is 1024MB with four 256MB DIMMs, but there is no "4th" memory slot. This user manual explains about you can install RAM memory up to 1024MB, but The Retro Web thinks the maximum memory you can get is 768MB, which might be true.

I may correct them soon, though.

PR #6777, opened 2026-02-03
Query UI font from system on Windows

aubymori:
Summary
=======
Previously, we determined the UI font from a predefined set of fonts each mapped to a language. This works well if the user doesn't change their Windows UI font, but if one does, they will get the default UI font associated with their language instead of the one the set.

This commit replaces ProgSettings::getFontName with ProgSettings::getUIFont, which uses the SystemParametersInfo API to query the message font from the system, which will allow users to have a custom font. It will also not interfere with different languages, as the message font will be appropriately set by default there.

Before the patch:
image

After the patch:
image

Also confirmed to work on hi-DPI:
image

Checklist
=========

### DIFF
--- a/src/qt/qt_main.cpp
+++ b/src/qt/qt_main.cpp
@@ -622,7 +622,7 @@ main(int argc, char *argv[])
     fprintf(stderr, "Qt: version %s, platform \"%s\"\n", qVersion(), QApplication::platformName().toUtf8().data());
     ProgSettings::loadTranslators(&app);
 #ifdef Q_OS_WINDOWS
-    QApplication::setFont(QFont(ProgSettings::getFontName(lang_id), 9));
+    QApplication::setFont(ProgSettings::getUIFont());
     SetCurrentProcessExplicitAppUserModelID(L"86Box.86Box");
 #endif
 
--- a/src/qt/qt_mainwindow.cpp
+++ b/src/qt/qt_mainwindow.cpp
@@ -2431,7 +2431,7 @@ MainWindow::changeEvent(QEvent *event)
 #ifdef Q_OS_WINDOWS
     if (event->type() == QEvent::LanguageChange) {
         auto size = this->centralWidget()->size();
-        QApplication::setFont(QFont(ProgSettings::getFontName(lang_id), 9));
+        QApplication::setFont(ProgSettings::getUIFont());
         QApplication::processEvents();
         main_window->centralWidget()->setFixedSize(size);
         QApplication::processEvents();
--- a/src/qt/qt_progsettings.cpp
+++ b/src/qt/qt_progsettings.cpp
@@ -28,6 +28,8 @@
 #ifdef Q_OS_WINDOWS
 #    include <QSysInfo>
 #    include <QVersionNumber>
+#    define  WIN32_LEAN_AND_MEAN
+#    include <windows.h>
 #endif
 
 extern "C" {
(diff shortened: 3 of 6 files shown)


OBattler:
imagem

Now it uses MS Gothic for Japanese, which is absolutely ugly.

OBattler:
And it also uses ugly fonts for both variants of Chinese as well, while Korean is somehow better.

aubymori:
> imagem
> 
> Now it uses MS Gothic for Japanese, which is absolutely ugly.

Seems to be using Yu Gothic UI for me after installing Japanese language pack.

image

I can also confirm that the message font is indeed set to that:

image

aubymori:
Here's the tool I am using to view and change the font, BTW. So you can check for yourself. https://gitlab.com/ftortoriello/WinClassicThemeConfig

kawapure:
It will use Yu Gothic UI on a Japanese system, but fall back to MS Gothic on English systems when using the Japanese locale because the message font will not be set accordingly. This isn't a particularly rare pattern. As a matter of fact, I don't think other Japanese fonts are installed on non-Japanese systems, as Japanese text always renders in MS Gothic in web browsers for me on my pure-English installations, and only uses other fonts when I install the Japanese language pack or install Japanese-language versions of Windows.

aubymori:
> It will use Yu Gothic UI on a Japanese system, but fall back to MS Gothic on English systems when using the Japanese locale because the message font will not be set accordingly. This isn't a particularly rare pattern. As a matter of fact, I don't think other Japanese fonts are installed on non-Japanese systems, as Japanese text always renders in MS Gothic in web browsers for me on my pure-English installations, and only uses other fonts when I install the Japanese language pack or install Japanese-language versions of Windows.

Oh, true. I hadn't considered that this person might be using 86Box in Japanese on an English system.

But yes, with the proper Windows language packs installed, it should use the correct font for the language just fine, given the rest of the OS will be using it too.

OBattler:
I'm quite sure I have the other Japanese fonts installed on my systems, in fact, I see Meiryo UI installed in Fonts, as is Yu Gothic UI, so the fonts are there, but some reason, Windows is defaulting to MS Gothic.

kawapure:
By the way, I feel that this is a better change, because I use Windows in Japanese with ClearType disabled and MS UI Gothic preferred as the message font. I will build this and test it on a clean Japanese-language installation to see.

aubymori:
> I'm quite sure I have the other Japanese fonts installed on my systems, in fact, I see Meiryo UI installed in Fonts, as is Yu Gothic UI, so the fonts are there, but some reason, Windows is defualting to MS Ghotic.

Could you use the utility I linked and check the font that "Message Box" is set to? It's probably a misconfiguration on your end...

OBattler:
It's most likely set to Segoe UI, given that I have my Windows language set to Portuguese (Portugal).

OBattler:
Also, your utility under executable files lists Windows 10 as the latest, I use Windows 11.

aubymori:
> Also, your utility under executable files lists Windows 10 as the latest, I use Windows 11.

Doesn't matter. It will work regardless.

kawapure:
Clean Japanese Windows 10 install versus my XP theme. I think it chooses the font well.

image

OBattler:
imagem

Segoe UI confirmed.

OBattler:
Just committed a workaround for these specific situations.

aubymori:
> Just committed a workaround for these specific situations.

No, this is bad. Now anyone using those languages will be forced onto those fonts. IMO, it's better to have no workaround. Using 86Box in an East Asian language on a system that isn't also using that language globally is such a weird edge case that I believe it's fine to leave how it is.

OBattler:
> No, this is bad. Now anyone using those languages will be forced onto those fonts.

They will only be forced if they use an East Asian language in 86Box but not for the OS itself, ie. in the weird edge case.

aubymori:
Sorry, I misread.

PR #6789, opened 2026-02-08
Qt: Fix media file picker location persistence

asemarafa:
Summary
  This PR fixes an issue where media file pickers (Floppy, CD, etc.) would reset to the 86Box home folder every time
  they were opened. This was particularly annoying during disk swaps for multi-disk installations, as it forced users to
  navigate back to their images directory repeatedly.

  Changes:
   - Updated MediaMenu::getMediaOpenDirectory() to return the home folder only on the first invocation. Subsequent calls
     return an empty string, allowing the native file dialog to persist its last visited location.
   - Updated MediaMenu::cdromMount to use getMediaOpenDirectory() for consistency across all media pickers.

  Checklist
   * [ ] Closes #xxx

  References
  N/A


OBattler:
Finally! This has been annoying me for ages.

PR #6918, opened 2026-03-12
Media History: stored write-protected media removed when machine starts

bozoscum:
Summary
=======
write-protected media other than which located in the machine folder being removed by Media History Manager due to 'wp://' is not positioned at the beginning of stored path in the machine configuration file.

eg. stored as ../../wp://disk/floppy.img instead of wp://../../disk/floppy.img

all the codes expect the beginning 5 character to be 'wp://' for write-protected but not in the middle of the path string

Checklist
=========
* [ ] Closes #xxx

References
==========
N/A

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -3311,8 +3311,6 @@ save_image_file(char *cat, char *var, char *src)
     else
         sprintf(temp, "%s%s", prefix, src);
 
-    fprintf(stderr, "temp: %s\n", temp);
-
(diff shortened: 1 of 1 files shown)


OBattler:
Because that's the intented usage, it's a prefix, similar to http:// or ftp://, not something intended to be used in the middle of a path.

OBattler:
Wait, I just noticed that I messed up in the implementation and you corrected it, thanks!

PR #6960, opened 2026-03-22
Add support for raw devices

tjemg:
Summary
=======
This PR allows to use a physical device as an HDD disk.
I have tested with an external CompactFlash card (2GB) and running under MacOS and under FreeBSD.
This is the first time I contribute something to the project which touches C/C++ code - therefore, I am surely missing lots of things here...
Note - I touched the CMakeLists.txt file to compile the software under MacOS with brew - maybe these changes can be removed?

What could be done better?
Currently I have tested this by selecting the "new..." under the HDD dialog - the CHS should be automatically chosen and shown, not only when clicking "ok".
When the device is detected to be a raw/block device, probably some settings should be greyed out in the disk dialog.
I have added a setting to the configuration to signal that the device is raw/block device.
Maybe this could also be made visible in the GUI?

Note: due to what I wrote above, this patch is most likely incomplete - while the bare functionality seems to work (and it kind of does what I want :-) ), it probably needs some further adjustments before it goes to the main branch; my intention of doing the PR is to show a bit of code to achieve an initial draft of the feature I would like to kindly ask to be implemented.

Thank you for the awesome project :-)

Checklist
=========
* [ ] Closes #xxx

References
==========

### DIFF
--- a/src/qt/qt_platform.cpp
+++ b/src/qt/qt_platform.cpp
@@ -78,6 +78,10 @@
 #endif
 
 #ifdef Q_OS_WINDOWS
+#    ifndef NOMINMAX
+#        define NOMINMAX
+#    endif
+#    include <windows.h>
 #    include <winioctl.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


OBattler:
It appears to be fine, except that windows.h should be included before winioctl.h in qt_platform.cpp, otherwise compilation is broken.

tjemg:
> It appears to be fine, except that windows.h should be included before winioctl.h in qt_platform.cpp, otherwise compilation is broken.

Thank you for pointing this out. I made a small change to reflect the comment.

PR #7109, opened 2026-05-02
SDL: New OSD and Shader Support

ChrisDeadman:
Summary
=======
I want to use 86Box on a Raspberry Pi without any X or Wayland, just boot into minimal init and 86box with SDL.
Since I was used to CRT shader on QT build, I was not happy to find out shaders are not supported in SDL backend, so this PR adds that shader support .

I've also implemented a new OSD based on imgui since I find the existing one to be a bit lacking.

Shader support is off by default, since it is not compatible with MacOS.
SDL Shaders can be enabled with `-DUSE_SDL_SHADER=ON`.
The old OSD is still the default, the new one can be enabled with `-DUSE_IMGUI=ON`.

image

image

image

Checklist
=========
* [x] Closes N/A
* [x] I have discussed this with core contributors already [(link to discussion)](https://github.com/86Box/86Box/discussions/7087)

### DIFF
--- a/.github/workflows/cmake_linux.yml
+++ b/.github/workflows/cmake_linux.yml
@@ -55,11 +55,15 @@ jobs:
           - name: SDL GUI
             qt: off
             qt6: off
+            sdl_shader: on
+            imgui: off
             slug: -SDL
             static: on
           - name: Qt 5 GUI
             qt: on
             qt6: off
+            sdl_shader: off
+            imgui: off
             slug: -Qt5
             packages: >-
               qtbase5-dev
@@ -71,6 +75,8 @@ jobs:
           - name: Qt 6 GUI
             qt: on
             qt6: on
+            sdl_shader: off
+            imgui: off
             slug: -Qt6
             packages: >-
               qt6-base-dev
@@ -104,6 +110,7 @@ jobs:
           build-essential
           ninja-build
           libfreetype-dev
+          libgles2-mesa-dev
           libsdl2-dev
           libpng-dev
           libc6-dev
@@ -127,6 +134,8 @@ jobs:
           -D NEW_DYNAREC=${{ matrix.dynarec.new }}
           -D QT=${{ matrix.ui.qt }}
           -D USE_QT6=${{ matrix.ui.qt6 }}
+          -D USE_SDL_SHADER=${{ matrix.ui.sdl_shader }}
+          -D USE_IMGUI=${{ matrix.ui.imgui }}
 
       - name: Build
         run: |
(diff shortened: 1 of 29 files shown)


OBattler:
I think it would be a good idea to remove the old OSD as well.

ChrisDeadman:
> I think it would be a good idea to remove the old OSD as well.

A few hours ago I would have agreed with you, but since the new OSD needs OpenGL it doesn't work on MacOS 🫤
At least I think so, don't really have experience with MacOS and do not own one.

OBattler:
Good point.

By the way, the shader code appears to break compile on MacOS:
```
FAILED: [code=1] src/unix/CMakeFiles/ui.dir/unix_sdl_shader.c.o 
/usr/bin/cc -DCMAKE -DDEBUG -DDISCORD -DUSE_DYNAREC -DUSE_SDL_UI -D_FILE_OFFSET_BITS=64 -I/usr/local/include/freetype2 -I/Users/runner/work/86Box/86Box/build/src/include -I/Users/runner/work/86Box/86Box/src/include -I/Users/runner/work/86Box/86Box/src/cpu -I/Users/runner/work/86Box/86Box/src/codegen -I/usr/local/Cellar/sdl2/2.32.10/include -I/usr/local/Cellar/sdl2/2.32.10/include/SDL2 -m64 -march=x86-64 -msse2 -mfpmath=sse -mstackrealign -fomit-frame-pointer -Wall -fno-strict-aliasing -Werror=implicit-int -Werror=implicit-function-declaration -Werror=int-conversion -Werror=strict-prototypes -Werror=old-style-definition  -ggdb -Og -std=gnu11 -mmacosx-version-min=10.13 -MD -MT src/unix/CMakeFiles/ui.dir/unix_sdl_shader.c.o -MF src/unix/CMakeFiles/ui.dir/unix_sdl_shader.c.o.d -o src/unix/CMakeFiles/ui.dir/unix_sdl_shader.c.o -c /Users/runner/work/86Box/86Box/src/unix/unix_sdl_shader.c
/Users/runner/work/86Box/86Box/src/unix/unix_sdl_shader.c:6:10: fatal error: 'GLES2/gl2.h' file not found
    6 | #include <GLES2/gl2.h>
      |          ^~~~~~~~~~~~~
1 error generated.
```

ChrisDeadman:
> By the way, the shader code appears to break compile on MacOS

Yeah that's exactly the thing, can't get the shaders to compile because GLESv2 is missing.
I've adapted the Linux script to include `libgles2-mesa-dev`, and enabled `USE_SDL_SHADER` there.
But no idea how to pull in GLESv2 on MacOS or if it is even supported.

OBattler:
Then I agree, the old OSD should be kept in as well.

PR #7110, opened 2026-05-02
Add configurable VM debug snapshots

spark2k06:
Summary
=======

Add a paused-VM debug snapshot facility for collecting emulator state into timestamped directories under the VM path.

The snapshot contents are configurable from the UI and currently include:

* CPU registers and bytes around CS:IP
* Conventional RAM dump, capped at 640 KB and reduced when the configured machine RAM is smaller
* Recent I/O port trace
* Active device list
* Detailed device dumps through an optional `device_t` callback

Detailed device snapshot support is currently implemented for:

* IBM EGA and compatible EGA variants (`ega.txt`, `ega_vram.bin`)
* IBM CGA and Pravetz VDC-2 (`cga.txt`, `cga_vram.bin`)

This is intended to make paused emulator state easier to inspect and share when debugging hardware-specific issues, while keeping the mechanism extensible for additional devices.

Checklist
=========

* [x] Closes N/A
* [x] This pull request does not require changes to the ROM set
* [x] This pull request does not require changes to the asset set

References
==========

The initial use case for this feature is comparing paused emulator state while improving compatibility in MiSTer PC/XT-family FPGA cores and derived projects. The snapshot output makes it easier to compare CPU state, conventional RAM, I/O activity, and video adapter state against 86Box.

Related projects:

* https://github.com/MiSTer-devel/PCXT_MiSTer
* https://github.com/MiSTer-devel/Tandy1000_MiSTer
* https://github.com/MiSTer-devel/PCjr_MiSTer

No ROM or asset changes are required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,34 @@ Building
 ---------
 For instructions on how to build 86Box from source, see the [build guide](https://86box.readthedocs.io/en/latest/dev/buildguide.html).
 
+VM debug snapshots
+------------------
+
+This development branch includes a VM debug snapshot facility intended to make paused emulator state easier to inspect and share during debugging.
+
+When the VM is paused, use the VM debug snapshot action to create a timestamped directory under the VM path:
+
+```
+debug_snapshots/snapshot_YYYYMMDD_HHMMSS/
+```
+
+The snapshot contents can be configured from the snapshot contents dialog. The currently selectable core data is:
+
+* CPU registers and bytes around CS:IP
+* Conventional RAM dump, capped at 640 KB and reduced automatically when the configured machine RAM is smaller
+* Recent I/O port trace
+* Active device list
+* Detailed dumps for active devices that expose snapshot support
+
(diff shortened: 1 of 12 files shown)


spark2k06:
Addressed the CodeQL findings in `de4a54c96`.

Changes made:
- Replaced `localtime()` with `localtime_s()` / `localtime_r()`.
- Centralized snapshot file creation through `debug_snapshot_fopen_write()`.
- Restricted POSIX-created snapshot files to user read/write permissions.
- Sanitized device snapshot directory names before using them as path components.
- Replaced the new `strncpy()` uses with bounded `snprintf()` copies.

spark2k06:
This remaining CodeQL path-injection alert is on the centralized snapshot file helper.

The filename component is now restricted to fixed/sanitized names and POSIX files are created with `0600`. The remaining tainted part is the VM/user path itself, which is intentionally user-configurable because snapshots are expected to be written under the VM path.

I believe this is a false positive for this feature rather than an additional security issue.

PR #7111, opened 2026-05-02
cdrom_image_viso: Fix more non-Windows breakage

ChrisDeadman:
Summary
=======
The latest VISO commits broke that functionality on Linux.

Commit `198db451a` changed VISO root directory stat population to read from `plat_dir_t`'s current entry instead of calling `stat()` directly. macOS already exposed the base directory as the initial current entry (via `fgetattrlist`), but the POSIX backend pre-loaded the first *child* entry instead, causing `plat_dir_is_dir()` to return false for the root and aborting the VISO mount.

Fix: stat the base directory path before entering the readdir loop and expose it as a synthetic `"."` current entry. `plat_dir_read()` advances past it normally. `plat_dir_rewind()` and `plat_dir_count_children()` restore the same state. No change to Windows or macOS backends; no change to callers.

Checklist
=========
* [x] Closes N/A

### DIFF
--- a/src/include/86box/plat_dir.h
+++ b/src/include/86box/plat_dir.h
@@ -427,6 +427,7 @@ typedef struct {
     size_t         path_len;
     DIR           *find;
     struct dirent *data;
+    struct dirent  root_data;
     struct stat    stats;
     uint8_t        stats_valid;
 } plat_dir_t;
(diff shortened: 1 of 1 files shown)


bhjdhgcehgrehge:
Screenshot_20260502_225532
It's fixed.

PR #7142, opened 2026-05-16
Added option "Auto-capture mouse"

usergithub64:
Added option "Auto-capture mouse"

Summary
=======
Added the "Auto-capture mouse" option to the Action menu with enabling/disabling the checkbox and saving this setting in 86Box.cfg for each VM. The "Auto-capture mouse" option is disabled by default. Added translation of this option to ru-RU.po and updated 86Box.pot. Tested when running the VM in a windowed mode, switching to fullscreen mode and back, with "Toggle UI in fullscreen", when running the VM immediately in fullscreen mode (86Box.exe -f).

Checklist
=========
* [ ] Closes #xxx
* [ ] I have tested my changes locally and validated that the functionality works as intended
* [ ] I have discussed this with core contributors already
* [ ] This pull request requires changes to the ROM set
  * [ ] I have opened a roms pull request - https://github.com/86Box/roms/pull/changeme/
* [ ] This pull request requires changes to the asset set
  * [ ] I have opened an assets pull request - https://github.com/86Box/assets/pull/changeme/

References
==========
_Provide links to datasheets or other documentation that helped you implement this pull request._

PR #7145, opened 2026-05-17
Mouse capture improvements

lemondrops:
Summary
=======
* Always capture mouse when entering full-screen mode;
* Fix a bug where the cursor reappears if the machine is hard reset while mouse capture is on;
* Remember the mouse capture state when a modal window is opened, and restore it once the window is closed.

Checklist
=========
* [x] Addresses https://github.com/86Box/86Box/discussions/6226#discussioncomment-16905614, closes #7142

References
==========
N/A

### DIFF
--- a/src/qt/qt_mainwindow.cpp
+++ b/src/qt/qt_mainwindow.cpp
@@ -1487,6 +1487,8 @@ MainWindow::on_actionFullscreen_triggered()
             emit resizeContents(vid_resize == 2 ? fixed_size_x : monitors[0].mon_scrnsz_x, vid_resize == 2 ? fixed_size_y : monitors[0].mon_scrnsz_y);
         }
     } else {
+        if ((mouse_type != MOUSE_TYPE_NONE) || machine_has_mouse())
+            emit setMouseCapture(true);
         video_fullscreen = 1;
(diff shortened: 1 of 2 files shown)


usergithub64:
In fullscreen mode with a hard reset, mouse auto-capture does not occur. Can this be fixed? In my PR it was already done that during soft reset and hard reset in windowed/fullscreen mode, with the "Auto-capture mouse" checkbox enabled, auto-capture of the mouse also occurred.

lemondrops:
> In fullscreen mode with a hard reset, mouse auto-capture does not occur.

Have you tested this with this PR specifically?

lemondrops:
I've managed to reproduce the hard reset bug - it also happens in windowed mode.

usergithub64:
> > In fullscreen mode with a hard reset, mouse auto-capture does not occur.
> 
> Have you tested this with this PR specifically?

yes

lemondrops:
@usergithub64 Try now, this should handle both the bug with the cursor reappearing after a hard reset (while the emulator still thinks that capture is enabled) and remembering/restoring the capture state when a modal window (such as a hard reset confirmation dialog) is open/closed.

dhrdlicka (approved):

dhrdlicka (approved):

PR #7153, opened 2026-05-18
FEATURE: Allow Audio sample rate to be set at the VM level via config…

nstarke:
…uration.  Pull list of sample rate options from currently selected soundcard out.

I used codex to provide the locale translations, looks like it picked up a few additional strings that were not yet translated.  If thats a problem I can submit this PR without the locale implementations, just let me know.


OBattler:
Merged into the version 6.1 branch because 6.0 is currently feature frozen.

PR #7205, opened 2026-05-31
Add OSD to Qt

ChrisDeadman:
Summary
=======
This PR cleans up the OSD to have a backend-neutral core and adds it to Qt build.
The default keybind for Qt is `CTRL+ALT+O`.

image

Checklist
=========
* [x] Closes N/A


src/qt/qt_mainwindow.cpp
```diff
@@ -1532,6 +1533,38 @@ MainWindow::FindAcceleratorSeq(const char *name)
...
+        auto *ke = static_cast<QKeyEvent *>(event);
+        if (ke->key() == Qt::Key_F11 && (ke->modifiers() & Qt::ControlModifier)
+            && osd_right_ctrl_down && (QApplication::activeWindow() == this || qt_osd_is_visible())) {
+            /* Accept so Qt delivers a normal key-press instead of firing the
+             * screenshot shortcut; the key handler above then toggles the OSD. */
+            event->accept();
+            return true;
+        }
+    }
+
```
dhrdlicka: Please add a customizable key binding for this instead of hardcoding RCtrl+F11 (which conflicts with the default bind for screenshots as you already mention in the comment)

dhrdlicka (changes requested):

src/qt/qt_mainwindow.cpp
ChrisDeadman: Ah, good catch, forgot to clean that up. Will do!
Regarding your osd-explorer PR, I guess it's best if I wait until that is merged and rebase on that.

Cacodemon345:
Does this work on Vulkan yet?

ChrisDeadman:
> Does this work on Vulkan yet?

No, it doesn't show on Vulkan yet, only OpenGL and Software atm.
But I will add it before un-drafting it.

src/qt/qt_mainwindow.cpp
ChrisDeadman: Fixed in ~`350da95`~ 0a15863

dhrdlicka: That's better, though I wonder - is there any particular reason why you opted not to follow the existing pattern of defining a `QAction`, binding it in `updateShortcuts` for windowed mode and handling it in the big `if (event->type() == QEvent::KeyPress && video_fullscreen != 0) {` block in `eventFilter` for fullscreen mode?

ChrisDeadman: I wasn't familiar enough with the Qt code...I've fixed it up now, I hope it's good now.

ChrisDeadman:
@Cacodemon345 
> Does this work on Vulkan yet?

I have added Vulkan support now, let me know if there's something  I should change.

Cacodemon345:
How complete is this by the way?

ChrisDeadman:
> How complete is this by the way?

With the vulkan-renderer in it's complete now - OSD is working in all renderers. I don't plan to do any changes unless there are requests.

Cacodemon345:
Any particular reason this doesn't use ImGui's Vulkan renderer?

ChrisDeadman:
> Any particular reason this doesn't use ImGui's Vulkan renderer?

Yeah, it didn't seem worthwhile adding since it expects raw handles (render pass, queue family, descriptor pool, etc.) that `QVulkanWindow` abstracts. Since we already need the rasterizer for software path, I found just reusing it was providing the least amount of friction.

OBattler:
Can you please apply the changes requested by @dhrdlicka ?

ChrisDeadman:
> Can you please apply the changes requested by @dhrdlicka ?

I already have, I guess he did not yet have time to look at it.

dhrdlicka (approved):

OBattler:
It appears to now spam this when using the OpenGL renderer:
imagem
. Can this be fixed somehow?

ChrisDeadman:
I will see that I reproduce this on windows and will fix.

ChrisDeadman:
> It appears to now spam this when using the OpenGL renderer

I see it happening on linux aswell when switching from vulkan backend to OpenGL. It's not happening when OpenGL is used at startup so I overlooked that :/

ChrisDeadman:
@OBattler PR #7236 fixes it for me under Linux.

PR #7249, opened 2026-06-06
Fix text EDID dumps being rejected

lemondrops:
Summary
=======
Fix text EDID dumps (edid-decode output) being rejected as too large, and show an error message if trying to load malformed dumps.

Checklist
=========
* [ ] Closes #xxx

References
==========
N/A


Cacodemon345 (approved):
Seems to look OK.

PR #7260, opened 2026-06-10
qt: Make automatic dialog box pauses user-configurable

chungy:
Summary
=======
The automatic pause for dialog boxes (including file selectors, sound gain, reset confirmation) is made user configurable here, especially if stopping the VM’s time is undesired, and that can be critical for certain server tasks.

Translatable UI strings aren't included here, and I'm open for the menu item to be renamed anyway.

Checklist
=========
* [ ] Closes #xxx

### DIFF
--- a/src/qt/qt_mainwindow.cpp
+++ b/src/qt/qt_mainwindow.cpp
@@ -775,6 +775,9 @@ MainWindow::MainWindow(QWidget *parent)
     if (do_auto_pause > 0) {
         ui->actionAuto_pause->setChecked(true);
     }
+    if (do_auto_diag_pause > 0) {
+        ui->actionAutoDiag_pause->setChecked(true);
+    }
     if (force_constant_mouse > 0) {
         ui->actionUpdate_mouse_every_CPU_frame->setChecked(true);
     }
@@ -1671,7 +1674,7 @@ MainWindow::eventFilter(QObject *receiver, QEvent *event)
         }
     }
 
-    if (receiver == this) {
+    if (receiver == this && do_auto_diag_pause > 0) {
         static auto curdopause = dopause;
         if (event->type() == QEvent::WindowBlocked) {
             window_blocked = true;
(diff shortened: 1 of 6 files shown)


src/qt/qt_mainwindow.cpp
```diff
@@ -775,6 +775,9 @@ MainWindow::MainWindow(QWidget *parent)
     if (do_auto_pause > 0) {
         ui->actionAuto_pause->setChecked(true);
     }
+    if (do_image_auto_pause > 1) {
```
lemondrops: Is this a typo?
```suggestion
    if (do_image_auto_pause > 0) {
```

src/qt/qt_mainwindow.cpp
```diff
@@ -1671,7 +1674,7 @@ MainWindow::eventFilter(QObject *receiver, QEvent *event)
         }
     }
 
-    if (receiver == this) {
+    if (receiver == this && do_image_auto_pause > 0) {
```
lemondrops: I believe this is going to affect more than just image dialogs...

lemondrops (changes requested):

src/qt/qt_mainwindow.cpp
chungy: Yes, it most likely is. :)

src/qt/qt_mainwindow.cpp
chungy: Possibly. The settings dialog doesn't seem affected at least.

lemondrops: That's because the settings dialog pauses the VM on its own (https://github.com/86Box/86Box/blob/master/src/qt/qt_mainwindow.cpp#L1274). Some of the dialogs affected would be the reset/exit confirmation ones, as well as miscellaneous ones like sound gain, "specify dimensions", etc (and possibly the preferences dialog too).

lemondrops: After an internal discussion, we've decided you can rename it to something like "Auto-pause in dialogs" to reflect what it does (settings dialog will however remain untouched to prevent issues when hard-resetting to apply the machine configuration). I'll later move it (alongside the other auto-pause option) to the preferences dialog and global config.

chungy: Oh, that shifts the feature's scope.  That could be fine I suppose.

chungy:
I've pushed a rebased and squashed commit with @lemondrops's suggestion to change the option's scope from file selectors to dialog boxes in general.  The settings dialog still does its own thing.

lemondrops (approved):

PR #7280, opened 2026-06-13
Rewritten Vulkan renderer with librashader support

Cacodemon345:
Summary
=======
Rewrites the Vulkan renderer with librashader support.

Checklist
=========

References
==========
None.


CMakeLists.txt
```diff
@@ -200,6 +200,17 @@ if(QT)
...
     endif()
+
+    if(APPLE)
+        option(LIBRASHADER "Build with librashader support" OFF)
+    else()
+        option(LIBRASHADER "Build with librashader support" ON)
+    endif()
+
+    if(LIBRASHADER)
+        add_subdirectory(corrosion)
```
dhrdlicka: As I already mentioned on Discord, librashader themselves recommend building it as a DLL when using it from a C codebase. Also personally I don't like the complexity that this adds to the build process

src/qt/qt_slangp.hpp
dhrdlicka, line 1: Is the filename supposed to have a dash instead of an underscore?

src/include/librashader_ld.h
dhrdlicka, line 1: Move 3rd party headers to `src/include` please

dhrdlicka (changes requested):

src/qt/qt_slangp.hpp
Cacodemon345: Yes.

src/include/librashader_ld.h
Cacodemon345: Done.

src/qt/qt_slangp.hpp
dhrdlicka: Why?

Cacodemon345: Consistency with the `qt-glslp-parser.h` header file name.

lemondrops: If anything, it's qt-glslp-parser.h that needs to be renamed.

PR #7304, opened 2026-06-18
Fix GDBSTUB not working when machine is started from manager

NyanCatTW1:
Summary
=======
GDBSTUB was broken when a machine is started from the VMManager (see discussion in https://github.com/86Box/86Box/discussions/7090)
The root cause was `gdbstub_init` getting invoked regardless of whether a machine is being run, which made VMManager listen to the stub port before any machine can start  
This PR fixes the issue by making sure `gdbstub_init` is only invoked if a machine is actually being started. The call is moved to `machine_init`, which is only invoked by `pc_reset_hard_init` while a machine is starting up

Checklist
=========
* [X] Closes issue mentioned in https://github.com/86Box/86Box/discussions/7090#discussioncomment-16792646

References
==========
Claude Opus 4.8 assisted in discovering the root cause & initial patch, I then refined the fix to its current state

### DIFF
--- a/src/86box.c
+++ b/src/86box.c
@@ -1363,8 +1363,6 @@ usage:
     if (lang_init)
         lang_id = lang_init;
 
-    gdbstub_init();
-
     /* All good! */
     return 1;
 }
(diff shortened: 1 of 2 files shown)


NyanCatTW1:
CC: @seadra @srmeister  
Would be great if you can test whether the fix also allows you to debug a machine started from UI

PR #7311, opened 2026-06-19
Implement AST RAMpage/XT ISA expanded memory (EMS)

skiretic:
Summary
=======
Makes the AST RAMpage/XT provide working LIM EMS. The board previously
registered but never delivered expanded memory to a driver — it did nothing in
practice; this implements the page-frame and window-register behavior the AST
memory managers expect.

This is an initial release of the board working in any capacity. It has so far
been tested against a single machine/driver configuration and needs further
testing across other configurations before it can be considered complete.

Implementation (all gated on `board == ISAMEM_RAMPAGEXT_CARD`; other isamem
boards are unchanged):

- The window odd byte is treated as the page-frame base register
  (`frame_addr = (val & 0x7f) << 14`), and only window 0 moves the frame. REMM
  relocates the frame through it while sizing the board — parking it over the
  card's own RAM to RAM-test, then committing the real segment — so honoring it
  keeps the frame reading as open bus during REMM's empty-64K-block scan.
- The window 1/2 odd readbacks report the top of conventional RAM
  (`D = min(mem_size, 640) >> 4`) in REMM's bank encoding, plus an identity
  latch, so REMM accepts the board and locates where the EMS region begins.
- The even byte maps EMS pages 0-based above conventional RAM (`index = page - D`)
  and echoes the raw page for REMM's write/read-back identity probe.
- The page windows are added with `mem_mapping` flag 0 instead of
  `MEM_MAPPING_EXTERNAL` so they route wherever REMM commits the frame (the
  lowest free 64K UMA block, not a fixed 0xD0000) regardless of that region's
  mem state, matching the hardware.
- A per-window mapping cache skips redundant `mem_mapping_recalc` while REMM
  sweeps the frame base across UMA, and `rampage_reset` darkens the windows on a
  warm reset so a stale mapping cannot survive a reboot.
- The start address now defaults to 0 (no backfill) so the board comes up as
  pure EMS instead of spending its RAM as conventional/extended memory.

The CMake change promotes `ISAMEM_RAMPAGE` out of `DEV_BRANCH` so the card
builds in regular (non-dev) builds.

Validated end to end on the one configuration tested: AST REMM 3.00 and REX 1.11
both dedicate the full configured 512K pool, and MS-DOS MEM reports expanded
memory under REMM 4.31. Builds warning-free with DEV_BRANCH off.

The reverse engineering and code in this pull request were produced primarily
using AI. This has also been disclosed separately and discussed with the dev
team outside of the content of this pull request message.

Checklist
=========
* [ ] Closes #xxx
  * [ ] I have opened an assets pull request

References
==========
- LIM/EMS 4.0 Expanded Memory Specification (Intel 300275-005, Oct 1987): https://www.phatcode.net/res/218/files/limems40.txt
- AST SuperPak Installation Supplement (REMM / REX drivers, 000347-001B, Jun 1987): https://vtda.org/docs/computing/AST/000347-001B_SuperPakInstallationSupplement_Jun87.pdf
- AST REMM.SYS 4.31 expanded memory driver (Internet Archive): https://archive.org/details/REMM431_ZIP

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -177,6 +177,8 @@ if(CMAKE_SYSTEM_NAME STREQUAL "OpenBSD")
     SET(CMAKE_EXE_LINKER_FLAGS "-Wl,-z,wxneeded")
 endif()
 
+option(ISAMEM_RAMPAGE "AST Rampage" ON)
+
 # Development branch features
 #
 #                      Option         Description                                Def.  Condition     Otherwise
@@ -184,7 +186,6 @@ endif()
 cmake_dependent_option(AMD_K5         "AMD K5"                                   ON    "DEV_BRANCH"  OFF)
 cmake_dependent_option(CDROM_MITSUMI  "Mitsumi CDROM"                            ON    "DEV_BRANCH"  OFF)
 cmake_dependent_option(G100           "Matrox Productiva G100"                   ON    "DEV_BRANCH"  OFF)
-cmake_dependent_option(ISAMEM_RAMPAGE "AST Rampage"                              ON    "DEV_BRANCH"  OFF)
 cmake_dependent_option(ISAMEM_IAB     "Intel Above Board"                        ON    "DEV_BRANCH"  OFF)
 cmake_dependent_option(ISAMEM_BRAT    "BocaRAM/AT"                               ON    "DEV_BRANCH"  OFF)
 cmake_dependent_option(OPL4ML         "OPL4-ML daughterboard"                    ON    "DEV_BRANCH"  OFF)
--- a/src/device/isamem.c
+++ b/src/device/isamem.c
@@ -135,8 +135,14 @@
 typedef struct emsreg_t {
     int8_t        enabled; /* 1=ENABLED */
     uint8_t       page;    /* page# in EMS RAM */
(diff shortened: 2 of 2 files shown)


Fenix770:
It is intresting!

I would be nice, for some time in the future if the documentation is available, to have the backfill option. Because on this card the backiflling was meant not only to backfiill the memory to 640kb but also to provide the large page frame feature.

(a drawback of this technique was it disabled any mainboard memory beyond 256kb, conventional and extended)
(I suspect this card also supports the EEMS spec, but I'm not sure, if is true it also can work without a 64kb page frame and also used with Deskview multitasker, like 386 machines can do).

I don't think this card would backfill extended RAM because it is an 8-bit XT memory board.
Also the Rampage 286 would not provide extended memory because the company sold a different extended memory card as separate product for OS/2 users.

Instead it provided an “extended memory emulator”  (REX.SYS) that only intercepted the int 15 calls and redirected them
to int 67 (the LIM EMS driver). 
There is a similar software on QEMM v6 and QRAM v1/2.

One program that needed this emulation was VDISK.SYS included in PC-DOS 3.0 to 3.3, because it suported only conventional or extended memory. 

Thanks in advance!

PR #7342, opened 2026-06-24
feat(isamem): implement AST RAMpage/XT EMS (LIM 3.2 + EEMS + backfill)

skiretic:
Summary
=======

Implements the **AST RAMpage/XT** expanded-memory card (8-bit ISA, PC/XT), replacing the `DEV_BRANCH`-gated stub in `isamem.c` with a working register-level model that is faithful across the real REMM driver lineage on genuine XT hardware.

What it does:
- **2 MB EMS/EEMS paging** — page-number / odd-companion register file, 64 mapping registers, UMA page frame, software `/START` frame relocation (C000/D000/E000).
- **Split Memory Addressing** — conventional-on-board backfill for starting address < 640 KB, driven by the `B|0x4001` switch byte (cap-encoder yields 64K granularity at every 64K-step start); EMS index = board page − D, backfill pages mapped as low system RAM.
- **Driver detection + per-version sizing** — page-port echo-latch detection; sizing quirks for REMM 2.08/2.70/3.00/3.10/3.30/4.10/4.70 resolved from driver RE. LIM-4.0-only drivers (4.30/4.31) faithfully fail (the /XT is LIM 3.2 + EEMS, not LIM 4.0).
- **Config UI** — three dropdowns (base I/O, memory size, starting address); switch bits derived internally.
- Ungates `ISAMEM_RAMPAGE` so the card builds in normal configurations.

Verified live on `ibmxt86` (genuine IBM XT 5160), screenshot-confirmed across the full size range (64K–2MB) and all nine starting addresses; conventional backfill validated by in-guest read/write into the reserved region, EMS validated by INT 67h map + read/write.

**Required XT PPI fix (`kbc_xt.c`):** REMM's memory-sizing parity check reads port 0x61 and tests bits 6/7. Those are 8255 Port B output bits (keyboard clock/clear) that read back 0 on real XT hardware (parity status is on Port C / 0x62), but `kbd_read` was returning them set — so REMM saw a false parity error and reported 0 EMS pages. Masking with `pb & 0x3f` (matching the canonical `port_6x.c` PPI) is required for the card to size correctly. It's a general XT-PPI correctness fix and can be split into its own commit if preferred.

Checklist
=========

References
==========
- AST RAMpage/XT User's Manual (000412-001 B, Jan 1987) — Split Memory Addressing, switch tables 2-2/2-3, Appendix E "How Rampage Works", 64-register paging.
- AST SuperPak User's Manual (000300-001 B, Oct 1985) — REMM/REX driver install.
- LIM EMS 3.2 specification; AST Enhanced EMS (EEMS) specification.
- REMM.SYS driver versions 2.08–4.70 (detection/sizing behavior reverse-engineered).

---

**AI disclosure:** developed with heavy AI assistance (reverse-engineering and implementation), under my direction and verified by me on real-hardware-faithful emulation. Discussed with core devs beforehand.


Fenix770:
If I recall correctly the Readme files of the different REMM drivers said that vesion with pair numbrers ( 4.30 for exmaple) were for the EMS cards  and not pair (like 4.31) were for mainboards AST 286 chipsets.

skiretic:
There is an error in my testing which lead to a false conclusion, I need to go backand recheck everything. Leaving as draft for now until confirmation.

Regardless, a lot more works now than did before. I just may have overlooked something, leading to the error in testing, thus false conclusion.

skiretic:
I am going to close this and do some further testing and changes.

PR #7351, opened 2026-06-27
EMU8K reset and performance improvements

ChrisDeadman:
Summary
=======
* [sound: EMU8K: move buffer reset to period boundary](https://github.com/86Box/86Box/commit/8e0456a160f3310b16423c27178ebbbeb902294b)
* [sound: EMU8K: avoid work when no voices are active](https://github.com/86Box/86Box/commit/392bafd06434c9a519b21d1f338c238eea398d07)

Checklist
=========
* [x] Closes N/A


OBattler:
I just reverted this because unreal9010 reports that the game Ultimate Race Pro hangs because of the changes introduced here.

Issue #7203, opened 2026-05-31
Mouse capture is broken in Gnome + Wayland compositor


amatas:
### Checklist

- [x] I **have searched** the issue tracker and **was unable** to find an [open](../issues?q=is%3Aissue+is%3Aopen) or [closed](../issues?q=is%3Aissue+is%3Aclosed) issue matching what I'm seeing.
- [x] I **have verified** that the issue is happening in the **[latest nightly build](https://ci.86box.net/job/86Box/lastSuccessfulBuild/artifact/)**, or the latest commit, if building from source.

### What happened?

The mouse capture works if the system pointer remains in the 86Box emulated window. If I move quickly the mouse to the top (as shown in the video) or outside the window in general the click events are lost inside the emulated OS.

I have attached a video of the behavior. The mouse system remains in the window and I can click on the widgets of the bios. If I press Ctrl+End the system cursor remains inside the window, in the second try I move quickly the mouse to the top, after that, the mouse clicks are missing inside the OS. When I press the Ctrl+end the system cursor is outside the main window at the top. (I click left and right buttons just to visualize the clicks in the recorder).

https://github.com/user-attachments/assets/7f6fae23-2f0c-4c72-9173-af497b828527

_I tried to troubleshoot this behavior by myself since many prior versions, but the implementation of the Wayland capture looks right and I'm a beginner with CPP and this compositor APIs._

Thanks for the great work!

### Configuration file

```ini
[S3 Vision964 PCI (ELSA Winner 2000 Pro/X)]
memory = 8

[General]
confirm_exit = 0
confirm_reset = 0
dpi_scale = 0
force_43 = 1
host_cpu = AMD Ryzen 9 9950X3D 16-Core Processor
scale = 3
uuid = 5fc8fdf4-9fdc-5ea9-ae45-b05a07af749a
vid_renderer = qt_vulkan
video_fullscreen_first = 0
video_fullscreen_scale = 0
video_gl_shader = D:/Games/86Box 3.1/shaders/dotmask.glsl
video_gl_vsync = 1

[Machine]
cpu_family = pentium_p55c
cpu_multi = 3
cpu_speed = 200000000
cpu_use_dynarec = 1
fpu_type = internal
machine = p5mms98
mem_size = 262144

[Video]
gfxcard = s3_trio64_pci
voodoo = 1

[Input devices]
joystick_0_nr = 0
joystick_type = 2axis_6button
keyboard_type = keyboard_ps2
mouse_type = ps2

[...]

PR #7490, opened 2026-07-21
Fix: Mouse capture with Vulkan and OpenGL under Gnome with wayland

amatas:
Summary
=======
This code fixes the mouse capture under Gnome + Wayland + QT6.

### Problem

 Under Gnome with Wayland, mouse pointer constraints (via `zwp_pointer_constraints_v1_lock_pointer`) operate on
 `wl_surface` objects — and each `QWindow` has its own surface. When using OpenGL or Vulkan renderers, the actual
 rendering happens in a separate `QWindow` (`OpenGLRenderer` / `VulkanWindowRenderer`), but the mouse capture code was
 passing the main window's `windowHandle()` to the capture callback. Wayland then locked the wrong surface, and
 pointer constraints silently failed — the cursor escaped the window.

 ### Fix

 `OpenGLRenderer` and `VulkanWindowRenderer` both inherit from `QWindow` (via class OpenGLRenderer : public QWindow,
 public RendererCommon). The existing `rendererWindow` member already points to these objects, so a
 `dynamic_cast<QWindow*>(rendererWindow)` gives us the correct window for pointer capture.

 Added a small accessor in `RendererStack`:

 ```cpp
   QWindow *captureWindow() const { return dynamic_cast<QWindow *>(rendererWindow); }
 ```

 For software rendering, dynamic_cast returns nullptr (SoftwareRenderer inherits from QWidget, not QWindow), and
 the caller falls back to this->windowHandle() — preserving the original behaviour.

Checklist
=========
* [x] Closes [#7203](https://github.com/86Box/86Box/issues/7203)

### DIFF
--- a/src/qt/qt_mainwindow.cpp
+++ b/src/qt/qt_mainwindow.cpp
@@ -405,8 +405,10 @@ MainWindow::MainWindow(QWidget *parent)
         if (mouse_capture) {
             if (hook_enabled)
                 this->grabKeyboard();
-            if (ui->stackedWidget->mouse_capture_func)
-                ui->stackedWidget->mouse_capture_func(this->windowHandle());
+            if (ui->stackedWidget->mouse_capture_func) {
+                auto *win = ui->stackedWidget->captureWindow();
+                ui->stackedWidget->mouse_capture_func(win ? win : this->windowHandle());
+            }
         } else {
             this->releaseKeyboard();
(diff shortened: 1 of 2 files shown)


OBattler:
Reverted it because I have received reports that it made things at least on KDE.

amatas:
Hi, I'll take a look also in KDE.

PR #7496, opened 2026-07-22
Fix: Mouse capture with Vulkan and OpenGL under Gnome with wayland under QT6

amatas:
Same as https://github.com/86Box/86Box/pull/7490 but only available for QT6 to avoid issues that caused the revert https://github.com/86Box/86Box/pull/7493

Summary
=======

Checklist
=========
* [x] Closes https://github.com/86Box/86Box/issues/7203
* [x] I have tested my changes locally and validated that the functionality works as intended. QT5 uses the previous implementation and QT6 the new one.

References
==========

### DIFF
--- a/src/qt/qt_mainwindow.cpp
+++ b/src/qt/qt_mainwindow.cpp
@@ -405,8 +405,14 @@ MainWindow::MainWindow(QWidget *parent)
         if (mouse_capture) {
             if (hook_enabled)
                 this->grabKeyboard();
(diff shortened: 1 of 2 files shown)


Lili1228:
This still breaks Plasma 6.

parkerlreed:
Yeah mouse breaks through capture on Plasma 6 here as well (Arch)

Only solution is software rendering at the moment. 

https://www.youtube.com/watch?v=iQqJvKkc-BU

Issue #505, opened 2019-12-20
More 8086 systems


jvernet:
**Is your feature request related to a problem? Please describe.**
Not really. 8086 systems present in 86Box are all special (Amstrad, Tandy, AT&T).

**Describe the solution you'd like**
More 8086 system, like the Deskpro 8086, PS/2 Model 25 or 30, and some other.

jvernet:
More over, did you plan to add missing machine from pcem ?
I would like to help, but never succeeded to have a working build env....

OBattler:
1. At the very least I do want to eventually add the PS/2 Models 25 and 30-86, but I'd need some proper documentation about them.

2. A development environment should be easy to setup: download MSYS2, run it in MINGW32 mode, do pacman -Syuu, after upgrading all the packages, restart and do pacman -Syuu again, then install mingw-gcc, I forgot which package for the Windows headers, freetype2 (you will have to move some include directories around because MSYS2 messes them up), libpcap, ghostscript, SDL2, and I believe the DirectX SDK, snd you should be good to go.

jvernet:
Will try again MSYS2 stuff.
About missing models, at least in par with PCEM V14:

- [8088] ATARI PC3
- [8088]Compaq Portable Plus
- [8088]NCR PC4i
- [8088]VTEC Laser Turbo XT
- [8086]VTech Laser XT3
- [286]Bull Micral 45
- [286]COMPAQ Portable II
- [286]DELL SYSTEM 200
- [286] EPSON PC AX
- [286] EPSON PC AX2e
-  [286] TULIP AT Compact
- [386SX] ACER 386SX25/N
- [386SX] AMI 386sx Clone
- [386SX]Packard Bell Legend
- [386DX]Compaq Pro 386
- [386DX]EPSON PCAX3

tiseno100:
Do we have documentation and ofc the ROMs for these machines? As i see these are from PCem. Some of these machines like the ACER 386SX25/N will require their chipset to be emulated.

jvernet:
Yes, these system are present in PCem
About the dev MSYS2 env: no way. After installing everything, I'm facing error like "off64_t unknown type, do you mean _off64_t". I replaced off65_t by _off64_t, passing. I have bunch of warning like ftello64, fseek64, undeclared, and now
"
win/win.c:376:6: erreur: « source_hwnd » non déclaré (première utilisation dans cette fonction)
  376 |  if (source_hwnd)
      |      ^~~~~~~~~~~
[...]

PR #7538, opened 2026-07-29
Add IBM PS/2 Model 25 emulation

Avastrap2:
Summary
=======

Add hardware-focused emulation of the original color IBM Personal System/2 Model 25 (Type 8525-004) with an 8 MHz 8086.

The implementation includes:

* the documented 512/640 KB system-board RAM mapping and control register
* system-board chip-select and parallel-output control
* a Model 25 keyboard/pointing-device gate-array interface instead of an 8042
* the documented IRQ1 dispatch through interrupt 71h
* integrated MCGA text, CGA-compatible, 640x480 monochrome and 320x200x256 modes
* PS/2 Model 25/30 floppy-controller behavior
* integrated 8250 serial and parallel interfaces
* DMA verify, UART loopback and writable-floppy formatting behavior exercised by the original diagnostics

No internal fixed-disk controller is added. The real Model 25 used a separate fixed-disk controller/card through its dedicated connector or an ISA storage adapter.

The proposal and test details are in discussion #7537. The machine was also requested historically in issue #505.

Validation
==========

* Normal POST reaches the expected boot path without keyboard or video POST errors.
* Sustained keyboard input remains operational.
* IBM Model 25 Starter Disk v1.00 boots.
* Its system tests advance through 100, 200, 300, 500, 600, 900, 1100, 1200 and 8600 when the test-600 scratch-disk write-protect instructions are followed.
* All changed device, machine, floppy and video object-library targets compile locally without warnings.
* `git diff --check` passes.

The complete current upstream Qt build is independently blocked in `qt_vulkanwindowrenderer.cpp`, and the no-Qt Windows frontend is independently blocked in `sdl_plat_win.c`; neither failure is in this pull request's changed files.

Checklist
=========

  * [x] I have opened a roms pull request - https://github.com/86Box/roms/pull/511

References
==========

* [IBM Personal System/2 Model 25 Technical Reference, first edition, June 1987 (SA23-1202)](https://bitsavers.org/pdf/ibm/pc/ps2/PS2_Model_25_Technical_Reference_Jun87.pdf)
* [IBM PS/2 Model 25 information and original diagnostic media](https://www.ibmfiles.com/pages/ps2model25.htm)

Development was assisted by Codex. Hardware behavior was checked against the IBM technical reference, the original BIOS and the original Model 25 diagnostic software.


OBattler:
The *code* seems fine. But let me test it first before I give my final verdict.

OBattler:
The PS/2 version of Windows 1.04 shows this instead of the splash screen on the Model 25-8086:
imagem .

Avastrap2:
I traced the issue to mcga mode detection. windows 1.04 requests BIOS mode 6 using cga mode-control value 0x18, as documented. The emulator incorrectly required the graphics bit as well, causing it to fall back to text mode.
I corrected the mode selection.

Screenshot 2026-07-30 053312
Screenshot 2026-07-30 053724

OBattler:
The 320x200 mode works but is not horizontally doubled for some reason:
imagem .

OBattler:
There's also weird black lines:
imagem .

OBattler:
The line doubling I have a rough idea of how to fix, but not those black lines. But 70 Hz is a bit strange. Let me see if the clock control matches this:
> [15:07](https://youtu.be/jCCaobFpKWc?t=1060) - this is wrong. MCGA also switches the dot clock in software. The tech ref has it on page 1-50: Mode control (3d4h, index 10h), bit 4 is documented as "clock select". IBM only documents 1 = 25.175 MHz, and "forgot" to document 0 = 14.318 MHz. You are right about the monitor type jumper, but the jumpers are just read by the BIOS that then programs the MCGA accordingly.

OBattler:
It does not:
```C
    /*
     * The formatter always uses a 25.175 MHz oscillator.  Low-resolution
     * modes divide the dot clock by two but retain the same 31.5 kHz line
     * rate.
     */
    dev->disp_on_time  = (uint64_t) (25.4220457 * (double) TIMER_USEC);
    dev->disp_off_time = (uint64_t) (6.3555114 * (double) TIMER_USEC);
    ```
    
    I'm going to have to replace that with a call to a recalctimings() function that i have to write.

Issue #6971, opened 2026-03-25
Seemingly GDI Rendering issues with numerous cards


junglira:
### Checklist

- [x] I **have searched** the issue tracker and **was unable** to find an [open](../issues?q=is%3Aissue+is%3Aopen) or [closed](../issues?q=is%3Aissue+is%3Aclosed) issue matching what I'm seeing.
- [x] I **have verified** that the issue is happening in the **[latest nightly build](https://ci.86box.net/job/86Box/lastSuccessfulBuild/artifact/)**, or the latest commit, if building from source.

### What happened?

There seems to be some sort of GDI rendering issue going on causing a crash in Microsoft Bob under Windows 98E - though I'm sure this extends beyond just Bob, as Bob uses incredibly standard Windows/Microsoft processes and methods (as of the 3.1/95 era) being a first party product.

Virtually every graphics card I've tried from various Machs, to several S3s, to a few Cirrus all have a rendering issue that causes a crash in Microsoft Bob when rendering a List that is wider than the screen space.

However, the Voodoo3 cards do not have this crash and can continue running without issue... at full (and near full) acceleration settings. However when lowering the acceleration  settings in Window's dialog to notch 1 or 2 on the four notch slider, it will have with the same crash. Notch 3 and 4 don't cause the crash.

Although it won't crash with the Voodoo3 cards on 3/4 acceleration settings, there is a rendering issue instead where certain elements don't draw correctly. This is specifically noticeable in the "speech bubble triangle" where it will render either transparent or in incorrect/broken colors. Additionally in selecting a background for a room, it will often not render the background piece correctly showing it's a deeper issue than merely the dialog balloon's triangle.

Though when using one of the other graphics cards or setting the Voodoo3 card to acceleration 1/2 settings, the rendering issue does not show up - and instead is replaced with the crash.

Attached are screenshots showing the speech balloon correctly rendered in normal situations and incorrectly rendered with the Voodoo3 with acceleration.

Correct rendering of speech balloon triangle:
<img width="800" height="600" alt="Image" src="https://github.com/user-attachments/assets/00435f9a-1c03-4286-9686-4f022fae8925" />

Two variations of the incorrect speech balloon triangle renderings:
<img width="800" height="600" alt="Image" src="https://github.com/user-attachments/assets/2fdd1f1b-fb06-4462-800b-d01ebffa9fca" />
[...]

PR #7580, opened 2026-08-02
mem: Invalidate page-crossing code blocks on writes to their second page

skiretic:
Summary
=======
Fixes the MS Bob NOTEBOOK/HOME invalid page faults under the new dynarec
reported in #6971 (the crash half of that ticket; the Voodoo3 speech-balloon
rendering is a separate 2D issue, not addressed here).

`addwritelookup()` chooses between the tracked write path (`page_lookup`,
which sets dirty bits so stale code blocks get invalidated) and the untracked
fast path (`writelookup2`, raw host pointer). The gate only checks whether a
block *starts* in the target page (`page->block`) and ignores blocks whose
code crosses a page boundary and ends there (`page->block_2`). A page holding
only the tail of a page-crossing block is judged code-free, so writes to it
never set dirty bits and the spanning block is never invalidated.

MS Bob's blitter emits self-modifying code straddling a page boundary and
patches it per blit, mostly in the tail page. The stale block re-executed
with outdated instruction boundaries and read misaligned immediates from the
rewritten bytes (a neighboring 0x66 prefix absorbed into `add edi, imm32`),
producing the corrupted EDI in the crash dialog. This also explains the
reporter's acceleration-slider correlation: higher GDI acceleration offloads
the blit to the video card and sidesteps the self-modifying CPU blitter.

The fix adds `page->block_2` to the gate, in the new dynarec variant only.
The old recompiler's variant of the gate shares the same shape, but the
issue reports the crash does not reproduce under ODR, and arm64 hosts force
`NEW_DYNAREC` on so I cannot test that path; it is left untouched rather
than shipping an untested change.

Verified on macOS arm64: the crash was deterministic per Scrapbook open
before the fix and gone after, across repeated opens and boots. The
mechanism was confirmed before patching: a debug knob rejecting reuse of
only page-crossing cached blocks (block shapes unchanged) eliminated the
crash. Regression-tested with full Win98SE boots and a sustained
multi-hundred-MB in-guest file copy; no faults or visible slowdown.

Checklist
=========
* [ ] Closes #xxx (partially addresses #6971 - the dynarec crash; the
      rendering issue remains)

References
==========
_Root cause analysis above; no external documentation involved._

### DIFF
--- a/src/mem/mem.c
+++ b/src/mem/mem.c
@@ -606,11 +606,15 @@ addwritelookup(uint32_t virt, uint32_t phys)
         writelookup2[writelookup[writelnext]] = LOOKUP_INV;
     }
 
+    /* A page holds code not only when a block starts in it (block) but also
+       when a block crossing a page boundary ends in it (block_2). Writes to
+       such a page must take the tracked page_lookup path, or self-modifying
+       code in the second page never invalidates the spanning block. */
 #ifdef USE_NEW_DYNAREC
 #    ifdef USE_DYNAREC
-    if (pages[phys >> 12].block || (phys & ~0xfff) == recomp_page) {
+    if (pages[phys >> 12].block || pages[phys >> 12].block_2 || (phys & ~0xfff) == recomp_page) {
 #    else
-    if (pages[phys >> 12].block) {
+    if (pages[phys >> 12].block || pages[phys >> 12].block_2) {
 #    endif
 #else
 #    ifdef USE_DYNAREC


skiretic:
Amended the branch: the original commit also patched the old recompiler's variant of this gate "for symmetry." I had missed that the ticket already established the crash does not reproduce under ODR, and since arm64 hosts force NEW_DYNAREC on, I have no way to test the old recompiler path here. Shipping an untested change to code the issue doesn't implicate was wrong, so that hunk is dropped -- the PR now touches only the new dynarec variant, which is the path I root-caused and verified. The PR description has been edited to match.

PR #7593, opened 2026-08-03
Fix IBM PS/2 Model 25 diagnostic failures

Avastrap2:
## Summary

- add a Model 25-specific IBM DBA fixed-disk controller variant
- implement Read/Write Extended, 48-bit ECC correction, CRC/check-field handling, and diagnostic status behavior
- preserve full-track Read Verify sector counts and restrict the OS/2 workaround to format commands
- model the Model 25 NMI-enable latch and clear stale controller ISR status on reset

## Root cause

The Model 25 was using the PS/1 controller path without the DBA extended-sector and ECC behavior expected by IBM diagnostics. Read Verify also collapsed a 17-sector request to one sector, while a count-based format workaround prematurely completed non-format commands. Controller reset could then surface an older ISR error when diagnostics returned to the control program.

## Impact

IBM Model 25 Starter and Advanced Diagnostics can exercise the fixed disk, ECC, full-disk Read Verify, and module-exit reset paths without false 1701 failures. The PS/1 controller remains a separate device variant.

Checklist
=========

* [ ] Closes #xxx

### DIFF
--- a/src/device/kbc_ps2_m25.c
+++ b/src/device/kbc_ps2_m25.c
@@ -25,6 +25,7 @@
 #include <86box/io.h>
 #include <86box/keyboard.h>
 #include <86box/machine.h>
+#include <86box/nmi.h>
 #include <86box/pic.h>
 #include <86box/plat_unused.h>
 #include <86box/timer.h>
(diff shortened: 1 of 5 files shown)


OBattler:
I actually think that perhaps, the PS/1 Model 2011 and PS/2 Model 30-286 may use those changes as well, it's just that noone has ever run any diagnostics on it before, not even Waltje who originally wrote this back in 2017.

JoshRodd:
Strong agree this should be a different device variant; I am trying to get the Model 30 BIOS working and ran into the exact same kind of failures.

PR #7607, opened 2026-08-04
Adds support for the ABIT BE6-II Slot 1 motherboard.

Slawterr:
Adds support for the ABIT BE6-II Slot 1 motherboard.
Changes include:
ABIT BE6-II machine definition and BIOS selection
Intel 440BX/PIIX4 platform configuration
Onboard HighPoint HPT366 ATA/66 controller
Dual HPT366 IDE channels, PCI configuration, interrupts, and bus-master DMA
Support for the HPT366 option module embedded in the Award BIOS
Tested with Windows 2000. The motherboard boots successfully, detects HPT366 disks, and works with the matching HighPoint driver.


MaxwellS04:
> Onboard HighPoint HPT366 ATA/66 controller

There is a standalone version of HPT366 which can be found [here](https://theretroweb.com/expansioncards/s/abit-hotrod-66). It would be nice if it was added.

ExAv2:
Another working RAID controller??

Slawterr:
> Another working RAID controller??

The HPT366/370 on these boards functioned only as a ATA66/100 controller without RAID function.

OBattler:
I merged this but I'm going to have to look into this because initializing the HPT366's BIOS separately should not be necessary - the machine BIOS should be able to just detect the device and use the built-in BIOS.

Slawterr:
> I merged this but I'm going to have to look into this because initializing the HPT366's BIOS separately should not be necessary - the machine BIOS should be able to just detect the device and use the built-in BIOS.

I'll look at this further as I was also going to add in the 1.2 and 2.0 revision boards with the HPT370 controller.

unreal9010:
@Slawterr could you also take a look at my ABIT BX6 code? I'm pretty sure I screwed something up (likely with GPIO emulation) because the boot process is abnormally fast compared to most other boards. Thank you!

Slawterr:
@unreal9010 I just ran a couple VMs in comparison. Its fast, but not wildly fast compared to the GA-686BX. I ran the comparison with a matching config. At a glance, the BX6 BIOS defaults to only the Primary Master being on auto detect while the 686BX has all channels on Auto. Setting all 4 channels to AUTO on the BX6 "feels" similar to that of the 686BX which has those enabled by default. If I turn off the three other channels on the 686BX, that feels similar to the defaults on the BX6 also. 

Let me know if making a similar change alters your perception of it being too fast.

unreal9010:
> Let me know if making a similar change alters your perception of it being too fast.

BX6 is a lot faster than the real deal and also considerably faster than the emulated ASUS P2B. In fact, it's so fast I can't see the CD-ROM drive entry at the POST screen (unlike P2B) because everything is processed so fast. Any way to tame it?

Slawterr:
Do you have a boot trace for the actual hardware?

OBattler:
Why does this have two functions? Makes no sense to me.

OBattler:
OK, it's apparently correct, according to the datasheet.

OBattler:
Got the built-in BIOS to load! I had to make sure the BIOS BAR was masked properly (ie. only bits 31-17 and 0 writable).

unreal9010:
> Do you have a boot trace for the actual hardware?

@Slawterr no, but here are some examples showing how the board works, confirming that the POST-related activities run too fast in 86Box:
https://www.youtube.com/watch?v=8j7OVwUBI7M
https://www.youtube.com/watch?v=ddvjUlxatho

PR #7626, opened 2026-08-06
Add the Intel Inboard 386/PC accelerator card for the IBM XT

Mike1978uk:
The Inboard 386/PC is an 8088-socket-compatible 386SX/486-class CPU upgrade card for the IBM PC/XT (5150/5160). It plugs into the 8088 socket and reuses the stock XT's own BIOS ROM, single 8259 PIC, and single 8237 DMA controller, but adds its own software-controlled memory wait states, A20 gating (via ports 0x60/0xA0/0xA1 - a stock XT has no 8042), and BIOS ROM shadow/cache control (ports 0x670/0x674). No new ROM file is needed - it uses the existing IBM XT BIOS ROMs already in the ROM repository.

Modeling this card surfaced three places where 86Box's timing model implicitly assumes CPU classification (`is286`/`is486`) tracks the real motherboard hardware, which doesn't hold for an accelerator card:
- IN/OUT port cycles and the LOOP instruction were entirely unscaled by CPU speed (new `io_waitstates`/`reg_op_waitstates` globals, consumed in `x86_ops_io.h`/`x86_ops_jump.h`, both zero by default - no effect on any other machine).
- `pic.c`'s IMR-write timing and `dma.c`'s XT-vs-AT 8237 detection both key off `is286`, which misclassifies a genuine single-PIC/single-8237 XT board paired with a 286-or-higher accelerator CPU. Both get a device-opt-in override (`pic_set_force_xt_imr_timing()`, `dma_set_force_xt()`) that leaves every other machine's behavior unchanged.

Ported to 86Box from UniPCemu's `hardware/inboard.c` (SuperFury, 2019-2022), itself derived from disassembly of the real INBRDPC.SYS driver (Intel's own driver for the card).

**Testing:** builds clean with no new warnings; boots to the machine's BIOS POST and Windows/DOS in the emulator. AI-assisted (Claude) for the 86Box-side port and timing-model integration; I understand and reviewed the changes.

Closes the request in Mike1978uk/win95-intel-inboard-386pc#1.

### DIFF
--- a/src/cpu/cpu.c
+++ b/src/cpu/cpu.c
@@ -175,6 +175,23 @@ int cpu_prefetch_width;
 int cpu_mem_prefetch_cycles;
 int cpu_rom_prefetch_cycles;
 int cpu_waitstates;
+int io_waitstates; /* Extra cycles added to IN/OUT instruction timing (x86_ops_io.h) -
+                       cpu_waitstates only ever throttles memory bus cycles; this covers
+                       I/O port cycles, which real ISA-bus hardware paces at a fixed bus
+                       clock independent of an accelerator CPU's own speed. Set by
+                       devices like inboard386.c that need to compensate for a fast CPU
+                       core executing I/O-bound timing loops (BIOS self-tests etc.)
+                       written assuming genuine 8088-class per-instruction throughput. */
+int reg_op_waitstates; /* Extra cycles added to pure register/branch instruction timing
+                           (currently just LOOP, x86_ops_jump.h) - covers the case where a
+                           period BIOS/driver delay or self-test loop is built entirely from
+                           register-only instructions (no memory or I/O access at all), so
+                           neither cpu_waitstates nor io_waitstates apply. A 386+-class core
+                           executes these natively far faster per cycle than an 8088 even at
+                           a matched clock, so a pure LOOP-based delay calibrated for genuine
+                           8088 timing finishes far too quickly on an accelerated CPU unless
+                           this is compensated too. Set by inboard386.c the same way as
+                           io_waitstates. */
 int cpu_cache_int_enabled;
 int cpu_cache_ext_enabled;
 int cpu_flush_pending;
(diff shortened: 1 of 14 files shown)


QuantumByteRider:
HI, 
The BIOS gives error 101 at POST and the "(1988) i386SX" Machine entry is Duplicated

Mike1978uk:
I had Claude pull out all the hooks and things we were using for debug, 101 was an error I saw before. I'm going to be away for a few weeks no access to look at it. But should be able to resolve on return. It was tested locally perhaps there's something in the setup. 

Leave me to resolve on my return. 

Apologies I had asked Claude to run and check before submitting. Clearly a minor bug slipped through removing some of the things used to build on.

There is a copy of binary and things on my repo.

Mike

Mike1978uk:
@QuantumByteRider — back from vacation, and this is now resolved. Fix is up as #7749.

You were right that it was in the setup, but not in yours: the machine shared `ibmxt_config`,
whose default is a 1982-dated 5160 ROM, and `INBRDPC.SYS` (the card's own required driver)
genuinely cannot work with that revision — it checks a signature at `F000:E05B` that the 1982
ROMs don't carry. So the machine booted an incompatible BIOS out of the box, which is the 101.

What made it hard to spot from my side: I had added the 1986 ROM entries to `ibmxt_config`
locally, so a `bios = ibm5160_050986` line worked on my tree and was silently ignored on
everyone else's — not a valid option in the shared list, no warning, falls back to the 1982
default. That is entirely my fault, and it is why I could not reproduce your report initially.
The machine now has its own BIOS list with only the two compatible 1986 revisions, so the bad
ROMs are unselectable rather than merely discouraged.

Validating that turned up two more real defects, both also fixed in #7749: none of the POST
fix-ups ran on 386-class CPUs at all (`cpu_set()` routes those to `exec386_2386()`, and they
only existed in `exec386()`), and memory timing was double-throttled on those same parts. A
plain 386DX — the CPU this card was actually sold with — now works.

The duplicate `(1988) i386SX` entry you spotted was fixed in-tree in the meantime; I confirmed
only one remains on master.

Tested through to a Windows 95 desktop on 09MAY86 + Mach8, 10JAN86 + Mach8, i386DX/25 + Mach8,
and on default settings with no `bios=` line — your exact case. Thanks for the report and for
the detail in it; it pointed straight at the real problem, and sorry it took the delay.

PR #7632, opened 2026-08-07
Add ABIT BE6-II revisions 1.2 and 2.0 HPT370 support

Slawterr:
Add ABIT BE6-II revisions 1.2 and 2.0 HPT370 support


OBattler:
I already told you that the ROM is supposed to be automatically loaded by the machine BIOS from what it has built in. The fact it's not loading that and you have to do it manually indicates you're doing something wrong. See how I fixed it on the HPT366 and apply this here as well.

PR #7645, opened 2026-08-09
sorted machines by year

luna-cant-code:
Summary
=======
I've sorted the machine types by year, because it was out of order.

Checklist
=========
* [ ] Closes #xxx

References
==========
My calendar


OBattler:
The problem is that the years are wrong, they're based on the CPU's release, but all the 8086 machines we emulate are, in fact, later than than the 8088 ones.

And while the 386SX was indeed later than the 386DX, 386SX chipsets are often often used for 286 as well and 386DX chipsets are often used for 486 as well.

luna-cant-code:
Well, that's good to know! Didn't think about that, thank you for the insight :)

PR #7668, opened 2026-08-11
Fix native file dialog blocking folder selection

wavz217:
Summary
=======
On a KDE 6 environment, when QT uses the function `QFileDialog::getExistingDirectory` to prompt the user to **select a folder** (e.g. when trying to mount a folder as a CD-ROM drive), it actually invokes an 'Open File' prompt, entirely preventing the user from selecting a folder.

There are 3 calls to the `getExistingDirectory` fron the UI: one in the CD-ROM mount routine and two in the GUI VM manager. This fix appends the two `QFileDialog::ShowDirsOnly` and `QFileDialog::DontUseNativeDialog` flags to the calls in order to skip the incriminated system dialog and to use Qt's own implementation which doesn't enable this behavior when the host runs Linux.

Checklist
=========
* [x] Closes N/A

References
==========
https://doc.qt.io/qt-6/qfiledialog.html

### DIFF
--- a/src/qt/qt_mediamenu.cpp
+++ b/src/qt/qt_mediamenu.cpp
@@ -654,8 +654,13 @@ MediaMenu::cdromMount(int i, int dir, const QString &arg)
 
     if (dir > 1)
         filename = QString::asprintf(R"(ioctl://%s)", arg.toUtf8().data());
-    else if (dir == 1)
-        filename = QFileDialog::getExistingDirectory(parentWidget, QString(), getMediaOpenDirectory(), QFileDialog::ShowDirsOnly | QFileDialog::DontUseNativeDialog);
+    else if (dir == 1) {
+        QFileDialog::Options options = QFileDialog::ShowDirsOnly;
+#ifdef Q_OS_LINUX
+        options |= QFileDialog::DontUseNativeDialog;
+#endif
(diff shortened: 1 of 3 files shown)


lemondrops:
The `QFileDialog::DontUseNativeDialog` flag would affect macOS and Windows, which I'm not sure is desired behavior.

wavz217:
It would only affect those three dialogs (one of which is the only one that really matters, the CD mount one) and it still works cross-platform nonetheless as it's Qt.

lemondrops:
Yes, but we prefer native dialogs whenever possible.

lemondrops:
Is mere `QFileDialog::ShowDirsOnly` not enough?

wavz217:
I know. The native dialog just doesn't work in this case. A check for the host system would have been better, but I think that it'd be too over-engineered. I'd rather just have this until they fix the bug in Qt upstream.

As far as I know, ShowDirsOnly is automatically enabled when there are no flags.

lemondrops:
> I think that it'd be too over-engineered.

That's not an excuse to degrade the user experience on other OSes.

wavz217:
I'll add a check. It will spawn Qt's dialog only if it's on Linux.

wavz217:
Should be done.

lemondrops (approved):

PR #7672, opened 2026-08-12
IDE: Overlap image reads with emulated command delays

Avastrap2:
## Summary

allows host HDD image reads to overlap the existing emulated IDE seek and transfer delays.

The host read begins when a supported command is issued. Its result is consumed by the original IDE completion callback before any guest-visible state is updated.

No IDE timing constants are changed.

## Supported commands

The asynchronous path applies to HDD execution of:

- `READ SECTORS`
- `READ SECTORS NO RETRY`
- `READ MULTIPLE`
- `READ DMA`
- Alternate `READ DMA`

ATAPI and write commands are unchanged.

## Implementation

A persistent read worker is created for each configured IDE HDD.

When a supported command is issued:

1. The existing command validation calculates the sector range.
2. The host read is submitted to that drive's worker.
3. The existing emulated seek and transfer callback remains scheduled.
4. The callback waits for the host result if it is not already complete.
5. The callback performs the original task-file, DMA, status, and IRQ updates.

If no asynchronous operation is available, the callback uses the existing synchronous read path.

## Threading model

Only `hdd_image_read()` and its private request state execute on the worker.

The worker does not modify:

- IDE task-file registers.
- IDE status flags.
- DMA controller state.
- IRQ state.
- Emulated command timers.

Only one read can be pending for each drive. The existing IDE callback consumes the result before using the sector buffer.

## Reset and shutdown

Pending reads are drained during:

- Device reset.
- Drive reset.
- IDE board teardown.
- Application power-off.
- HDD image synchronization.

The worker is joined before its events, sector buffer, or HDD image can be released.

Checklist
=========

* [ ] Closes #xxx
  * [ ] I have opened an assets pull request
- [ ] I have opened a docs pull request

### DIFF
--- a/src/disk/hdc_ide.c
+++ b/src/disk/hdc_ide.c
@@ -45,6 +45,7 @@
 #include <86box/hdc_ide.h>
 #include <86box/hdd.h>
 #include <86box/rdisk.h>
+#include <86box/thread.h>
 #include <86box/version.h>
 
 /* Bits of 'atastat' */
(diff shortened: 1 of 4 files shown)


Cacodemon345:
Does OS/2 work correctly with your changes?

OBattler:
Warp 3.0 and 4.0 work.

PR #7677, opened 2026-08-12
Add selectable and usable Level2 cache amounts to the AH4T (SiS 85C471)

mw308:
Summary
=======
New functionality which adds a selectable L2 cache amount to the 85C471 chipset, specifically targeted at the Abit AH4T.  The code also allows the cache to be usable at the varying amounts; tested with cachechk.

It should be ok to extend out to the existing AH4 as well, and potentially other SiS 471 boards.  It also could be a starting place for L2 cache on other machines.

Tested both the AH4T for changes as well as the AH4 to ensure the changes do not follow.

Requires dynarec to be enabled to ensure 100% emulation speed.

Checklist
=========
* [ ] Closes #xxx

References
==========
Datasheets at https://theretroweb.com/motherboards/s/abit-ab-ah4t


lemondrops:
What is the performance impact of this?

TC1995:
Does performance drop in terms of %?

mw308:
86box performance had no noticeable drop, emulation remained at 100%. Emulated ram performance goes from 17 microseconds in cache to 23 microseconds per KB out of cache. Previously, all ram performance was 17 microseconds regardless of cache size.

mw308:
How can you rerun the checks? I think the one it failed on was due to a GitHub issue at the time

mw308:
Actually now that DOS is running there's a 10% reduction in emulated speed.  Leave that with me.

mw308:
Still bad in Windows

mw308:
Abandoned due to low performance in Windows, even with dynarec on.

PR #7708, opened 2026-08-16
Add support for Iomega Ditto parallel port drive

dbrant:
Summary
=======
This adds support for an Iomega Ditto tape drive, connected to an LPT port with EPP. The drive and cartridge are configurable through the usual LPT port "Configure" UI.

Ditto drives have read-write support for Ditto-capacity cartridges (2GB+), and also have read-only support for a wide range of smaller legacy cartridge capacities. This is confirmed working in the first-party Ditto Tools software for Windows 95/98. It also works in the standard Microsoft Backup in Windows 98 (Read/write with the "Iomega QIC-3020" drive, and read-only with "Ditto 2GB+" drive -- MS Backup explicitly doesn't support writing to newer Ditto drives). It also works with the same DOS-based HP Colorado Backup exec tool that supports the FDC variant.

A few caveats to consider:
* I'm conscious that this is quite a lot of code (in addition to my previous QIC-117 FDC drive), which might become a maintenance burden depending on how the project evolves. I won't be offended if you don't want to take this onboard; in which case an alternative might be to ditch the existing FDC drive and use this one instead, since it's arguably more versatile.
* Ditto cartridges always came _pre-formatted_, and there's actually no way for the end-user to format a truly blank cartridge, i.e. an empty or zeroed image. This means that users of the emulator will need to use pre-formatted [tape images](https://github.com/dbrant/ftape/tree/master/blanktapes), or existing tape dumps.
* The cartridge capacity must be selected explicitly in the configuration, and is not inferred from the tape image. It's important to make sure this selection matches the tape image being used.

Monitor_1_20260816-100051-004

Checklist
=========


OBattler:
You forgot to add it to the list in lpt.c.

OBattler:
Nevermind, RichardG told me it's now char.c.

sebastien64:
suplement information for "iomega ditto"

https://en.wikipedia.org/wiki/Ditto_(drive)

PR #7719, opened 2026-08-17
Make MEM_BLOCK_NR externally configurable

Valefungo:
Summary
=======
I'm creating a side project for 86box, a layer to build it bare metal on a Raspberry Pi Zero 2 and Pi 4 (and Pi 5 when I'll buy it...)

I need to be able to change the new dynarec size from an external makefile because a raspi zero cannot allocate 128 mb for it.

Since I'm working hard to avoid touching the 86box code for this, the only point where I can't do nothing is this one.

I don't want to disrupt anything, this should be invisibile for everyone.

Cheers,
Valerio


Cacodemon345:
This needs to be guarded behind a `static_assert` macro in one of the below functions.

PR #7755, opened 2026-08-24
Miscellaneous machine table and init changes

sentogahara:
Summary
=======
* Rename the "Packard Bell PB810" to "BCM FM530" and add the AST Bravo EL BIOS variants as a selectable option in the device configuration window
* Fix some mistakes with the TMC PAT45PV sections of machine.h
* Rename all instances of "Award Modular BIOS" and "Phoenix - AwardBIOS" to just "AwardBIOS" in the BIOS version selectors, to reduce the length of the device configuration windows 

Checklist
=========
* [ ] Closes #xxx
  * [x] I have opened a roms pull request - https://github.com/86Box/roms/pull/528

References
==========
https://theretroweb.com/motherboards/s/bcm-fm530

### DIFF
--- a/src/include/86box/machine.h
+++ b/src/include/86box/machine.h
@@ -1175,6 +1175,10 @@ extern int             machine_at_p55t2s_init(const machine_t *);
 extern int             machine_at_ap5vm_init(const machine_t *);
 extern int             machine_at_p55tvp4_init(const machine_t *);
 extern int             machine_at_5ivg_init(const machine_t *);
+#ifdef EMU_DEVICE_H
+extern const device_t  pb810_device;
+#endif
+extern int             machine_at_pb810_init(const machine_t *);
 extern int             machine_at_8500tvxa_init(const machine_t *);
 extern int             machine_at_presario2240_init(const machine_t *);
 extern int             machine_at_presario4500_init(const machine_t *);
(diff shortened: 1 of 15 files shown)


MattBytesRetro:
"Rename the "Packard Bell PB810" to "BCM FM530" and add the AST Bravo EL BIOS variants as a selectable option in the device configuration window."

This is incorrect IMO. The PB810 isn't identical to the FM530. The PB810 has a different video chip, the Trio64V2/DX, vs the FM530's Trio64V+. Also, the 810 has built in USB, the FM530 does not. This can be confirmed by digging into the BIOS' that I did with the Discord team, and the multiple PB810 boards I have.

I would suggest separating these into two separate entries, one for the PB810 with it's bios, and one for the FM530 with it's.

Also I have attached the latest known BIOS for the FM530 that adds large disk support (>8GB). This works on the PB810, BUT breaks the internal video chip (external video works just fine) - Hence also confirming they are different boards.

[FM530-UniCorev2.2.zip](https://github.com/user-attachments/files/31401325/FM530-UniCorev2.2.zip)

sentogahara:
Actually, they are identical. BCM just offered multiple variants of the same board with different video chips and USB count, which are common board variant options offered to different OEMs with different costs. In fact, even AST seemingly used at least three different variants; while the Bravo EL variant I added has the Trio64V2/DX, the Adventure 2xx/4xx variant instead uses Trio32/64/64V+, and the Advantage 2000 and Adventure 4400 variants uses S3 ViRGE/DX/GX even. Since we can only emulate the Trio64V2/DX variant due to machine table limitations for now, I only added the Bravo EL BIOS.

The UniCore BIOS in question does not work with onboard video because it was made for another variant of the same board with a different video chip. Otherwise, it is fully compatible with this board.

PR #7780, opened 2026-08-26
Input: Use Ctrl+Alt+G to release mouse

mihainadas:
Summary
=======
Change the default mouse release shortcut from Ctrl+End to Ctrl+Alt+G in the Qt and SDL frontends. Ctrl+End is difficult to discover on compact keyboards because End may require a platform-specific Fn chord, and Qt renders it as a diagonal arrow glyph on macOS. Ctrl+Alt+G is mnemonic for toggling the input grab, avoids navigation keys, does not conflict with another default 86Box binding, and is also the convention used by QEMU. Custom user bindings are preserved.

This depends on #7779, which prevents the release shortcut trigger from reaching the guest before capture is released.

Checklist
=========
* [ ] Closes #xxx
* [ ] I have tested my changes locally and validated that the functionality works as intended
* [ ] I have discussed this with core contributors already
* [ ] This pull request requires changes to the ROM set
  * [ ] I have opened a roms pull request - https://github.com/86Box/roms/pull/changeme/
* [ ] This pull request requires changes to the asset set
  * [ ] I have opened an assets pull request - https://github.com/86Box/assets/pull/changeme/
* [ ] This pull request adds, changes or removes an external dependency
  * [ ] I have opened a docs pull request - https://github.com/86Box/docs/pull/changeme/

Testing
=======
Built the macOS Qt frontend successfully and compiled the changed SDL frontend translation units successfully.

References
==========
QEMU uses Ctrl+Alt+G to toggle mouse and keyboard grab: https://www.qemu.org/docs/master/system/keys.html

PR #7785, opened 2026-08-26
Input: Prevent menu activation on mouse release

mihainadas:
Summary
=======
Prevent the mouse release shortcut from activating a guest menu when its held modifiers are released. The trigger key is consumed before reaching the guest, but Ctrl and Alt have already been forwarded by the time the complete chord is recognized. Sending a break-only event for the trigger key before releasing the modifiers preserves the absence of a guest key press while preventing Windows from interpreting Alt as a standalone tap. Apply the same ordering to the Qt and SDL frontends.

Checklist
=========
* [ ] Closes #xxx

Testing
=======
Built the complete macOS Qt frontend from current master and compiled the changed SDL translation unit. Reproduced the regression in Windows 95 WordPad, then verified that Ctrl+Alt+G releases capture without entering text, moving guest content, or highlighting the File menu.

References
==========
Follow-up to #7779 and #7780. VirtualBox documents the same break-before-Alt-release technique for preventing unwanted Windows menu activation: https://www.virtualbox.org/ticket/2613

### DIFF
--- a/src/qt/qt_mainwindow.cpp
+++ b/src/qt/qt_mainwindow.cpp
@@ -1696,6 +1696,9 @@ MainWindow::eventFilter(QObject *receiver, QEvent *event)
         QKeyEvent *ke = (QKeyEvent *) event;
         if (mouse_capture) {
             if ((QKeySequence) (ke->key() | (ke->modifiers() & ~Qt::KeypadModifier)) == FindAcceleratorSeq("release_mouse") || (QKeySequence) (ke->key() | ke->modifiers()) == FindAcceleratorSeq("release_mouse")) {
+                /* Prevent an Alt-based shortcut from looking like a standalone
+                 * Alt press to the guest when the held modifiers are released. */
+                this->keyReleaseEvent(ke);
                 keyboard_all_up();
                 plat_mouse_capture(0);
                 event->accept();
--- a/src/unix/sdl_main.c
+++ b/src/unix/sdl_main.c
@@ -617,6 +617,9 @@ main(int argc, char **argv)
                             const int key_down = event.key.down;
 #endif
                             if (key_down && (keyboard_get_shift() & 0x11) && (keyboard_get_shift() & 0x44) && (xtkey == 0x22) && mouse_capture) {
+                                /* Prevent an Alt-based shortcut from looking like a standalone
+                                 * Alt press to the guest when the held modifiers are released. */
+                                keyboard_input(0, xtkey);
                                 keyboard_all_up();
                                 plat_mouse_capture(0);
                                 break;


OBattler:
By the way, perhaps you may want to join a Discord server where most of the development talk goes on.

mihainadas:
@OBattler, since you mentioned the Discord, there is one idea I’d really like your instinct on before I go too far down the rabbit hole.

I was born in 1987, and 86Box has hit a very specific nostalgia nerve for me. Growing up, I spent a lot of time reading about the hardware I couldn’t afford, first the faster Pentiums and later the Core 2 / GeForce 8800 era. Recreating those machines now is a big part of why I’ve started contributing.

What I’m wondering is whether there is a technically sane way to go past the current 86Box limits without trying to extend the existing low-level approach all the way. For something like a 2007 Core 2 + GeForce 8800 GTX machine, could a hybrid approach make sense? Keep the period-correct machine, BIOS, chipset and device model, but use dynamic translation for the CPU and some form of HLE for the GPU, ideally still exposing a convincing 8800 GTX to XP and translating G80 commands/shaders to Metal or Vulkan underneath.

The goal for me would not simply be “run old XP games”. There are already easier ways to do that. The fun would be installing XP, the original chipset drivers and ForceWare, seeing a GeForce 8800 GTX in Device Manager, running 3DMark06, and effectively recreating the dream PC I could only read about in 2007.

Does that basic direction sound technically sane to you? And do you think it preserves enough of what makes 86Box interesting, or does that level of HLE cross a line where it stops making sense for this project?

If you think there is something worth exploring there, I’d be very interested in trying to prototype parts of it.

OBattler:
The only way would be to somehow have a virtualizing CPU core for anything beyond Pentium II (or maybe even for Pentium II itself), but that would also require rewriting the device timer system to be synchronized to host time instead of the emulated CPU's TSC, which would in turn require a lot of the code to be made thread-safe that is currently not.

As for using GPU acceleration, the ATi Rage 128 emulation Patrons currently have exclusive access to (but it's going to master as soon as it's deemed to be bug-free enough) has that as an option using Vulkan/Metal, so that's doable.

But I think that for emulating a 2007 CPU and graphics card, some sort of QEMU fork may be the better option, at least for now. Maybe in 10 years when there's going to be something like the Apple M16 with 15000 single thread points, it would become feasible to fully emulate.

mihainadas:
@OBattler Thanks, that gives me a much clearer picture. I think I’ll poke at a few small prototypes before committing to one direction.

I’ll start with QEMU and see how close a mostly stock setup can get to XP on a plausible Core 2-era machine, including the original chipset drivers. Then I want to make a bare-minimum 8800 GTX PCI device, just enough to see how far ForceWare gets. If I can reach command submission, I’ll try capturing something tiny, like a clear or a D3D9 triangle, and see whether translating the G80 commands and shaders to Vulkan or Metal is realistic.

I’m also curious enough to try one small experiment in 86Box, using host time and an accelerated CPU loop while keeping the device work on one thread. I won’t try to turn that into a rewrite. It would just tell me whether the timing problem can be boxed in or whether QEMU is clearly the better home for this.

My guess is that I’ll end up with QEMU too, but I’d rather have a couple of concrete results to point at first. I’ll let you know what I find.

PR #7803, opened 2026-08-27
Add ability to load CD images into RAM

Cacodemon345:
Summary
=======
Add ability to load CD images into RAM.

Checklist
=========

References
==========
None.

### DIFF
--- a/src/86box.c
+++ b/src/86box.c
@@ -234,6 +234,8 @@ int      is_new_808x = 0;                                         /* (C) Use the
 
 int      gdbstub_port = 12345;                                    /* (C) The GDB stub port. */
 
+int      cdrom_ram_thread_enabled = 0;                            /* (C) Load CD images into RAM. */
+
 // Accelerator key array
 struct accelKey acc_keys[NUM_ACCELS];
 
--- a/src/cdrom/CMakeLists.txt
+++ b/src/cdrom/CMakeLists.txt
@@ -30,6 +30,7 @@ add_library(cdrom OBJECT
     cdrom_image_viso.c
     cdrom_mke.c
     cdrom_chd.c
+    cdrom_ram.c
 )
 target_link_libraries(cdrom chdr-static)
 if(NOT WIN32)
--- a/src/cdrom/cdrom.c
+++ b/src/cdrom/cdrom.c
@@ -21,6 +21,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <stdlib.h>
+#include <stdatomic.h>
 #include <wchar.h>
 #include <86box/86box.h>
 #include <86box/device.h>
@@ -42,6 +43,7 @@
 #include <86box/scsi_cdrom.h>
 #include <86box/sound.h>
 #include <86box/ui.h>
+#include <86box/thread.h>
 
 #define RAW_SECTOR_SIZE    2352
 
@@ -3370,6 +3372,49 @@ cdrom_update_status(cdrom_t *dev)
     }
 }
 
+struct thread_ram_load_status_t
+{
+    _Atomic(uint64_t) cntr;
+    atomic_bool_t     finished;
+};
+
+typedef struct thread_ram_load_status_t thread_ram_load_status_t;
+
+struct thread_ram_load_ctx_t
+{
+    void* cd_image_ctx;
+    uint8_t* ram_image_to_read_info;
+    const cdrom_ops_t* ops;
+    uint64_t begin_pos;
+    uint64_t sectors_to_read;
+
+    thread_ram_load_status_t* load_status;
+};
+
+typedef struct thread_ram_load_ctx_t thread_ram_load_ctx_t;
+
+void
+display_cntr(void* param)
+{
+    thread_ram_load_status_t* cntr = (thread_ram_load_status_t*)param;
+
+    while (!atomic_load(&cntr->finished)) {
+        plat_delay_ms(1);
+        ui_set_prog_dialog(atomic_load(&cntr->cntr));
+    }
+}
+
+void
+ram_load_thread(void* param)
+{
+    thread_ram_load_ctx_t* load_param = (thread_ram_load_ctx_t*)param;
+
+    for (uint64_t i = 0; i < load_param->sectors_to_read; i++) {
+        load_param->ops->read_sector(load_param->cd_image_ctx, &load_param->ram_image_to_read_info[(i + load_param->begin_pos) * 2448], i + load_param->begin_pos);
+        atomic_fetch_add_explicit(&load_param->load_status->cntr, 1, memory_order_relaxed);
+    }
+}
+
 int
 cdrom_load(cdrom_t *dev, const char *fn, const int skip_insert)
 {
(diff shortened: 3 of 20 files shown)


dhrdlicka:
I reiterate my doubts about the usefulness of this feature. If your disk I/O is a bottleneck for CD-ROM emulation at, let's say 8x-16x speed, then the problem is definitely not in 86Box.

Cacodemon345:
> I reiterate my doubts about the usefulness of this feature. If your disk I/O is a bottleneck for CD-ROM emulation at, let's say 8x-16x speed, then the problem is definitely not in 86Box.

This is for games doing heavy disk I/O on the CD/DVD drives (playing back movies, music, audio, or even loading assets). Descent III does it, also GTA III.

dhrdlicka:
Again, GTA3 requires a 4x drive and recommends a 8x one. What kind of a disk can't sustain 1200KB/s read?

Cacodemon345:
GTA3 isn't the only game doing it. Also a lot of people are stuck with slow 5400RPM HDDs on laptops that will not take kindly to heavy I/O operations.

dhrdlicka:
A healthy 5400RPM drive still has throughput in the neighborhood of 100MB/s. 8x speed CD-ROM reads definitely are not a heavy I/O operation

Cacodemon345:
My laptop disagrees. I/O operations on the 5400RPM HDD, combined with the CPU being Intel i5-8250U, is slow as hell. If Windows is also doing I/O operations on it at the same thing it will leave little bandwidth available for other applications including the emulator itself.

OBattler:
Again, I didn't have these problems on my old Pentium Dual-Core E5700 with a 5400 RPM 500 GB hard disk drive and only 4 GB of RAM so Windows would constantly page memory to and from it.

Cacodemon345:
Vista/7 was much better at handling HDDs too. Windows 10/11 is atrocious at handling HDDs.

OBattler:
I ran Windows 10 on that machine from mid-2015 onwards, still didn't have 86Box's CD-ROM image access lagging the emulation at all.

OBattler:
Oh and on my Windows 7 before that, I had an exaggerated number of fonts installed, as well as a crapton of other stuff, to the point it took ages to start, so it was hardly a healthy install of Windows 7. Yet, still no such problems.

Cacodemon345:
My 8.1 install despite being 32 bit barely ran on my old Pentium Dual Core PC either. I couldn't run anti virus programs on it at all without those wrecking up the system performance. All on a 1GB of RAM and a measly 160 GB Samsung HDD.

DDXofficial:
> My laptop disagrees. I/O operations on the 5400RPM HDD, combined with the CPU being Intel i5-8250U, is slow as hell. If Windows is also doing I/O operations on it at the same thing it will leave little bandwidth available for other applications including the emulator itself.

I have personally run Windows on 5400 RPM HDDs for an extended period of time in the past and can confirm I have never experienced any noticeable slowdowns while loading CD-ROM images into 86Box.

> Vista/7 was much better at handling HDDs too. Windows 10/11 is atrocious at handling HDDs.

Okay, and? I fail to see how relevant this is to the feature being added.

> My 8.1 install despite being 32 bit barely ran on my old Pentium Dual Core PC either. I couldn't run anti virus programs on it at all without those wrecking up the system performance. All on a 1GB of RAM and a measly 160 GB Samsung HDD.

This is GitHub, Caco. Please remain on topic.

Cacodemon345:
> I have personally run Windows on 5400 RPM HDDs for an extended period of time in the past and can confirm I have never experienced any noticeable slowdowns while loading CD-ROM images into 86Box.

I was testing Sonic R on a Windows 98 SE VM at some point on my laptop. I had to install WDM drivers at some point to avoid the audio playback bugs.

I enabled the digital audio playback option in Windows as I wasn't exactly fond of having the audio being played back through the drive subjected to various CD audio filters and the sound card mixers for various reasons.

I couldn't maintain 100% speed at all in the game with that thing enabled and the WDM drivers active. An emulated 66MHz Celeron was too slow for it. So I increased the speed, but that instead made the emulator itself much more sensitive to slowdowns because both the CD audio was being played back by Windows and the game itself was also being very taxing on the VM.

That's one of the reasons for this PR, to not have the CD emulation itself become the bottleneck on slower host machines at all, especially on more demanding games with the ability to play and load assets from the CD image.

OBattler:
> I enabled the digital audio playback option in Windows as I wasn't exactly fond of having the audio being played back through the drive subjected to various CD audio filters and the sound card mixers for various reasons.

But that's literally subject to the same filters and mixers.

> I couldn't maintain 100% speed at all in the game with that thing enabled and the WDM drivers active. An emulated 66MHz Celeron was too slow for it. So I increased the speed, but that instead made the emulator itself much more sensitive to slowdowns because both the CD audio was being played back by Windows and the game itself was also being very taxing on the VM.

I need to test that.

andrew-hoffman:
Wouldn't a read-ahead cache provide the same benefit without needing to keep the whole disc in RAM? Do the multiple loading threads help when they are all bottlenecked on 5400 rpm disk I/O?

Cacodemon345:
> Wouldn't a read-ahead cache provide the same benefit without needing to keep the whole disc in RAM? Do the multiple loading threads help when they are all bottlenecked on 5400 rpm disk I/O?

The images are loaded one time into the RAM and as fast as possible utilizing the maximum possible read speeds of the disk. Once the loading is done the disk is never accessed again.

Cacodemon345:
Yet another reason for this PR is to decompress Aaru format and MAME CHD images entirely into RAM. Unlike other formats where compression cannot be a source of slowdowns, decompression time can become an issue on slower host machines.

andrew-hoffman:
Preloading the whole disk might only take 10 seconds of loading/decoding delay for a CD image off a 5400 RPM hard drive, but then it would be more like 1 minute for a DVD image or 2 minutes for a DVD-9 or large CHD. That's doing a lot of unnecessary work if you don't listen to the whole CD or play the whole game in that session, and also taking up enough RAM that it's likely to get paged out before you need it and then you're back to having stutters.

If the decoding is done in blocks ahead of the read pointer (something on the order of 1-10mb, I'm not sure what size would be best or how many blocks to keep in cache but that could be tunable) then the CPU work is more spread out and there's less unnecessary work done and less of a risk of the memory that you need being paged out. Games expect optical drives to have some seek delay but then provide a steady data stream once it starts reading. 

It might also be possible to send better hints to the host OS that optical disc image files are going to be read sequentially and then the OS can manage the read ahead caching.

chungy:
I think this has dubious value also.  Operating system caching should take care of any disk-read performance issues that are being attempted to fix here; smarter file systems will even start preloading the file into cache when an application begins sequentially reading it (ZFS does this: it detects read patterns and tries to predict future requests).  On top of that, some images can be very large.  For "accurate" DVD drive images, up to 8.5 GB.  You can easily go bigger with unbounded drive images.  It's a pretty significant ask for the emulator process itself to eat up all that memory, with no option for the OS to release the memory under pressure (as it does under normal file caching operation).

---

On compression:

CHD was specifically designed so that MAME can use it in real time during its own emulation.  Each block in a CHD file is independently compressed (and can be compressed with different algorithms, even) and performance doesn't suffer much.  Use `chdman copy -c none -i compr.chd -o uncompr.chd` if you suspect compression is being a bottleneck.  I would personally be surprised to see practical benefit in 86Box from it.

AaruFormat, I don't really know the details of, but the `aaru` command itself can be fairly slow on some operations with compressed images.  No idea if that's a format limitation or implementation issue.  `aaru image convert -O compress=false compr.aaruf uncompr.aaruf` can likewise be used to test potential performance issues related to compression.

Cacodemon345:
Simple Google/DuckDuckGo searches for lists on which PC games run from CD turns up nothing. And I determined most can be installed to HDDs anyway.

PR #7810, opened 2026-08-28
bin/cue/ccd: Open all binary non-audio images on Windows with sequential access hint

Cacodemon345:
Summary
=======
bin/cue/ccd: Open all binary non-audio images on Windows with sequential access hint.

Checklist
=========
* [ ] I have tested my changes locally and validated that the functionality works as intended

References
==========
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=msvc-170.

### DIFF
--- a/src/cdrom/cdrom_image.c
+++ b/src/cdrom/cdrom_image.c
@@ -497,7 +497,11 @@ bin_init(const uint8_t id, const char *filename, int *error)
 
     memset(tf->fn, 0x00, sizeof(tf->fn));
     strncpy(tf->fn, filename, sizeof(tf->fn) - 1);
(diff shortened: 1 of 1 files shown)


dhrdlicka:
Shouldn't we do this for all optical images?

Cacodemon345:
The rest are fully abstracted, and not readable/writable with normal C standard I/O.

OBattler:
I'm quite sure this change also applies to ISO, MDS, and MDX.

Cacodemon345:
Those depend on the common backend for doing C standard file I/O, so those are fine.

PR #7849, opened 2026-09-03
Add Elo TouchSystems SmartSet serial touchscreen emulation

Xeon3D:
Summary
=======
Adds emulation of an Elo TouchSystems SmartSet-family serial touchscreen
controller (the E271-2200/E271-2210 controllers' native "10-byte" binary
protocol: 0x55 lead byte, 8 command/response bytes, checksum), as a new
device alongside the existing 3M MicroTouch serial touchscreen
(`mouse_microtouch_touchscreen.c`), which it's modeled on: same
`tablet_devices[]` slot, same serial-attach lifecycle, same
`mouse_get_abs_coords()`-driven polling, same NVRAM-backed calibration
persistence.

New file `src/device/mouse_elo_touchscreen.c`, wired into
`src/device/CMakeLists.txt`, `src/include/86box/mouse.h`, and
`src/device/mouse.c`'s `tablet_devices[]` — no UI-layer changes needed,
the device list is enumerated generically.

Implements: Acknowledge, Touch (stream + query), Mode (binary and ASCII
forms), Reset (soft/hard), Calibration (both the set-by-range and the
Offset/Numerator/Denominator packet forms, plus the controller-driven
two-point auto-calibrate sequence), Scaling (same two forms), Trim-mode
clamping, Parameter (baud rate reconfiguration), Quiet, ID, Jumpers,
Owner, and NVRAM save/restore.

A handful of secondary commands (Report, Filter, Timer, Key, Low Power,
and the Configuration/'g' full-state dump-restore) are present as
accept-and-echo stubs that store/return the value but have no other
functional effect — no driver exercised in testing (see below) depends
on them actually doing anything, but they're called out at the top of
the file as known-incomplete rather than silently pretending to be full.

Testing
=======
Built and tested in a Windows 98 SE 86Box VM against real period Elo
software (staged from the vendor's own driver/utility packages, not
written for this device):

* DOS: `COMDUMP.EXE`/`COMTEST.EXE` (raw protocol dump, no driver) show
  correctly framed and checksummed packets while dragging across the
  touch area.
* DOS: `ELODEV.EXE` v1.7d (the real Elo DOS TSR driver) detects the
  controller, passes power-on self-test, and reports correct
  identification (AccuTouch type, "EloInc." owner string, firmware
  2.1); `ELOCALIB.EXE` v1.6c completes a real 3-point calibration
  against it and the calibrated/scaled coordinates track correctly
  afterward (verified against known scaled ranges, e.g. touching the
  bottom-right corner in an 80x25-scaled mode reports exactly (80,25)).
* Windows 98 SE: the Elo Win9x MonitorMouse driver (v03.00.00) installs
  without any Plug-and-Play involvement (it isn't a PnP-aware device on
  real hardware either), and its Diagnostics panel reports "The
  touchscreen is installed and working properly."

References
==========
* Elo TouchSystems, "SmartSet Touchscreen Controller Family Technical
  Reference Manual" (1993) - vendor documentation for the E271-2200 /
  E271-2210 / E271-2201 / E271-2202 controllers and their command set.
* Cross-referenced against the Linux kernel's own implementation of the
  same wire format: https://github.com/torvalds/linux/blob/master/drivers/input/touchscreen/elo.c

Checklist
=========
* [x] I have tested my changes locally and validated that the
      functionality works as intended
* [ ] This pull request requires changes to the ROM set (n/a - no ROM)
* [ ] This pull request requires changes to the asset set (n/a)
* [ ] This pull request adds, changes or removes an external dependency (n/a)

### DIFF
--- a/docs/elo/00-plan.md
+++ b/docs/elo/00-plan.md
@@ -1,0 +1,128 @@
+# ELO Serial Touchscreen Emulation — Plan
+
+## Goal
+
+Add a new 86Box device emulating an Elo TouchSystems **SmartSet** serial
+touchscreen controller (the "10-byte packet" protocol — the modern/native Elo
+serial protocol, as opposed to the legacy 3/4/6-byte emulation formats also
+supported by the same silicon). This lets guest OSes running period-accurate
+Elo drivers (DOS, Windows 3.1x, Windows 9x) talk to a virtual touchscreen the
+same way they would to a real E271-2200/2210/2201/2202 SmartSet controller,
+using 86Box's existing absolute-pointer ("tablet") input plumbing.
+
+Model: 86Box already ships `mouse_microtouch_touchscreen.c`, a 3M MicroTouch
+serial touch device. It is the closest existing analog — same category
+(`tablet_devices[]`), same serial-port attach pattern, same NVRAM-backed
+calibration persistence, same `mouse_get_abs_coords()` input source. The ELO
+device follows the same file shape and lifecycle, with an Elo-specific command
+parser instead of MicroTouch's SOH/CR ASCII command language.
+
+## Why SmartSet ("10-byte") and not the legacy formats
+
+The Linux kernel's `drivers/input/touchscreen/elo.c` supports four wire
+formats keyed by a `proto` field: a 10-byte packet (its default/primary path),
+plus 6-byte, 4-byte and 3-byte legacy formats. Cross-referencing that with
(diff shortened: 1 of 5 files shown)


Xeon3D:
Withdrawing this one. It was written with AI assistance and I haven't reviewed it myself to the standard CONTRIBUTING asks for, so it shouldn't be sitting in your queue taking up reviewer time. Apologies for the noise.

PR #7818, opened 2026-08-30
Add IBM PC 730 and 750 (type 6877/6887) machine

Avastrap2:
## Summary

Adds emulation support for the IBM PC 730 and PC 750 (types 6877 and 6887), based on the Intel 82430FX Triton PCIset, with onboard S3 Trio64V+ graphics and National Semiconductor PC87306 Super I/O.

## Checklist

- [x] I have tested my changes locally and validated that the functionality works as intended.
- [ ] I have discussed this with core contributors already.
- [x] This pull request requires changes to the ROM set.
  - [x] I have opened 86Box/roms#533.
- [ ] This pull request requires changes to the asset set.
- [ ] This pull request adds, changes, or removes an external dependency.

## References

- IBM PC 700 Series types 6877/6887 Hardware Maintenance Manual
- Intel 82430FX PCIset datasheet
- Intel 82371FB PIIX datasheet and specification update
- National Semiconductor PC87306 Super I/O datasheet

PR #7867, opened 2026-09-05
Add revised IBM PC 730/750 (types 6877/6887) support

Avastrap2:
## Summary

I have reintroduces IBM PC 730/750 machine from #7818 with revised reset, configuration storage and power-management handling.

The shared fixes are separate commits

- invalidate instruction mappings when PAM changes the backing of the current page.
- clean up injected keyboard state when a reset interrupts Ctrl+Alt shortcuts.
- latch enabled PIIX/PIIX3 IRQ wake events and deliver pending requests when the SMI gate opens.
- preserve the first displayed character for the IBM VGA blanking rollover case without overriding other blank-end timings.
- expose fresh secondary NVRAM state so board defaults do not overwrite loaded data.

There is no BIOS-address-specific reset hook or deferred hard-reset workaround. IBM CMOS defaults are initialized by the board, not by the shared Super I/O implementation.

## Validation

I have tested with a Pentium 133 and 16 MB RAM configs

- BIOS setup save/exit with an existing configuration.
- Fresh configuration initialization, saving settings and a cold boot from the saved NVRAM.
- Windows 95 Restart twice, returning to the desktop.
- Restart in MS-DOS mode followed by Ctrl+Alt+Del, returning to Windows works.
- Windows 95 now completely shutdown.
- 3 bios APM standby/keyboard-wake cycles and two intervening reboots with APM available, using both dynamic recompilation and interpreter override.
- 13 S3s blanking regression cases,  +focused keyboard-state and PIIX IRQ-latch tests.

## Scope and limitations

Keyboard wake is guest-tested; IRQ3/4/8/12 have register-level checks but not guest wake tests. The fresh-saved configuration boots but reports AH=86h/CF=1 for the APM calls; it is not counted as a successful standby test. The board power-controller handshake remains a minimal model of the observed firmware protocol, not complete power-controller emulation.

CPU limited

- I have limits CPU to only Intel-based pentiums due to they doesn't know others CPU.

## Checklist

- [x] I have tested my changes locally and validated the functionality listed above.
- [ ] I have discussed this revision with core contributors already.
- [ ] This pull request requires new changes to the ROM set.
- [ ] This pull request requires changes to the asset set.
- [ ] This pull request adds, changes or removes an external dependency.

## References

- Previous PR: [86Box/86Box#7818](https://github.com/86Box/86Box/pull/7818).
- Intel 82371FB/82371SB datasheet, SMI control/enable/request registers: https://www.mouser.com/catalog/specsheets/intel%20corporation_29055002.pdf
- IBM PC 700 Series types 6877/6887 Hardware Maintenance Manual, publication 52H3381.

### DIFF
--- a/src/86box.c
+++ b/src/86box.c
@@ -1594,49 +1594,52 @@ pc_send_ca(uint16_t sc)
         /* Use R-Alt because PS/55 DOS and OS/2 assign L-Alt Kanji */
         keyboard_input(1, 0x1D);  /*  Ctrl key pressed */
         if (keyboard_get_in_reset())
-            return;
+            goto cleanup;
         keyboard_input(1, 0x138); /* R-Alt key pressed */
         if (keyboard_get_in_reset())
-            return;
+            goto cleanup;
         keyboard_input(1, sc);
         if (keyboard_get_in_reset())
-            return;
+            goto cleanup;
         usleep(50000);
         if (keyboard_get_in_reset())
-            return;
+            goto cleanup;
         keyboard_input(0, sc);
         if (keyboard_get_in_reset())
-            return;
+            goto cleanup;
         keyboard_input(0, 0x138); /* R-Alt key released */
         if (keyboard_get_in_reset())
-            return;
+            goto cleanup;
         keyboard_input(0, 0x1D);  /*  Ctrl key released */
         if (keyboard_get_in_reset())
-            return;
+            goto cleanup;
     } else {
         keyboard_input(1, 0x1D); /* Ctrl key pressed */
         if (keyboard_get_in_reset())
-            return;
+            goto cleanup;
         keyboard_input(1, 0x38); /* Alt key pressed */
         if (keyboard_get_in_reset())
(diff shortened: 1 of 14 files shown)


OBattler:
The changes to the Intel chipsets are going to break Award BIOS'es.

OBattler:
Because what the Award BIOS'es do is - toggle shadow RAM read but expect the previous code to continue executing. But I think this can be solved with a machine gate.

OBattler:
As for the SMI#, that is already handled by apm_set_do_smi:
```C
                    apm_set_do_smi(dev->apm, !!(val & 0x01) && !!(fregs[0xa2] & 0x80));
                    if ((val & 0x01) && !(old & 0x01) && fregs[0xaa])
                        smi_raise();
```.

OBattler:
Edit: Ahhh, pending IRQ's! That's actually a very good point here.

OBattler:
Merged and added the machine gate to the i4x0 change.

PR #7897, opened 2026-09-07
CPU: Fix debug/breakpoint handling in dynarec

lotharsm:
Summary
=======
_This was done using a Claude Opus session. Claude connected to the VM via GDB and performed testing. I monitored the VM, and I can say that it is visibly working. Since it touches emulation core, it needs review. It "works for me", but might break other stuff. Viewer's discretion is advised._

Yesterday, I was debugging Empire Earth, and found out that the game only works when the debug registers are enabled. I did this, and the game launched. As mentioned briefly yesterday, I ran into the issue that I could later NOT verify the results anymore.

The reason for this: GDBSTUB enabled forced 86Box to run in interpreter mode, and not with the dynarec engaged. When enabling the dynarec, the code path that triggers the debug breakpoints never triggered as soon as the VM even booted with dynarec, so flipping back to interpreter mode didn't help.

This is fixed by https://github.com/86Box/86Box/commit/dd804f8850dcb89ec7925fc9709ab1388f83a26c - the game launches, even if using the dynarec. I *manually* tested ODR and NDR. Having `-DDEBUGREGS486=ON` is STILL a requirement though.

https://github.com/86Box/86Box/commit/1f48eb29d386632961d58cba13a93dcaeedda5c8 and https://github.com/86Box/86Box/commit/a13b217a9907adb18490c10722b426feb10962e0 were reported as bugs by a sub-agent performing a review.

This had no visible outcome, and I couldn't see any regressions testing the game.

So, after the three first commits: **Empire Earth correctly launches using the dynarec*.

And finally: https://github.com/86Box/86Box/commit/1e72618991a212268c75c7726bc59fd1171b00c9

This is potentially the most "dangerous" one, but it lead to a _huge_ performance improvement. In Empire Earth as a test case, it performed almost as good with `-DDEBUGREGS486=OFF`, so the performance penalty introduced by the debug registers dropped by _a lot_. Probably needs more testing though...

Performance is not completely on par with the "regular" builds though, but my testing capabilities are a bit limited, given that I only have a Ryzen 5 3600 that tops out somewhere around 200 Mhz.

Checklist
=========
* [ ] Closes #xxx

References
==========
* Tested with GDB inside the VM, hooking into the game process
* Reasoning: https://gist.github.com/lotharsm/d0bb822f405641d8bbcd2cfd8c248a20
* _NO_ outside documentation!

### DIFF
--- a/src/cpu/386.c
+++ b/src/cpu/386.c
@@ -277,7 +277,12 @@ exec386_2386(int32_t cycs)
                 if (trap & 16)
                     dr[6] |= 0x2000;
                 trap = 0;
+                /* RF must be set in the EFLAGS image x86gen() pushes, so the
+                handler's IRET resumes the instruction instead of faulting on it
+                again; delivery itself leaves RF clear for the handler. */
+                cpu_state.eflags |= RF_FLAG;
                 x86gen();
+                cpu_state.eflags &= ~RF_FLAG;
                 /* No instructions executed at this point. */
                 break;
             } else if (cpu_16bitbus) {
--- a/src/cpu/386.c
+++ b/src/cpu/386.c
@@ -266,15 +266,20 @@ exec386_2386(int32_t cycs)
             if ((ol == 3) && opcode_has_modrm[fetchdat & 0xff] && (((fetchdat >> 14) & 0x03) == 0x03))
                 ol = 2;
 
-            if (is386)
-                ins_fetch_fault = cpu_386_check_instruction_fault();
-
-            /* Breakpoint fault has priority over other faults. */
-            if ((cpu_state.abrt == 0) & ins_fetch_fault) {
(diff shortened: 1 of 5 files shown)


lotharsm:
Just want to clarify that this is ONLY tested by behavior, I don't have a proper test suite for it.

Furthermore, I am *not* familiar with the inner workings of the CPU core itself.

Before merging or adapting for further investigation, this MUST be reviewed by one of the wizards.

OBattler:
This looks fine to me, just the two bug fixes should be applied to 386.c exe as well, because the bugs affect the 286/386 interpreter as well.

lotharsm:
And done. Had to force push a fix for an oversight, so the QA checks are currently complaining - let's see if this passes or if I confused CI.

PR #7932, opened 2026-09-11
floppy: fix raw-image track placement and stale track state

JoshRodd:
Summary
=======
Adds support for IBM double-stepped 360KiB 3.5" diskette formats, a 67.5 TPI format, used by upcoming IBM PC JX support. Some of this code was derived from existing double-step support for 5.25" drives.

If the machine type explicitly enables it, a 360KB image file will be presented as the IBM 67.5 TPI double stepped format. Low level images like .IMD, .86F, etc. are unaffected.

If not enabled, mounting a 360KB image file continues to fill in tracks 0-39 and leaves tracks 40-79 empty, presenting it as a 135 TPI diskette, per usual. This feature could also work on a 1.44MB drive, although no machine types are known or planned which would enable it.

This feature is essentially useless without a BIOS that supports it; a BIOS that doesn't support it will end up presenting 360KB images in an unusable way.

We retain the physical seek input when reloading after FORMAT, invalidate flux/turbo/sector views before an invalid seek returns, and reject out-of-range track and sector-buffer access, and keep the normal D86F logical-track convention.

AI Assistance
============
AI extensively used to research, design, implement, and test this.

Assisted-by: oh-my-pi:gpt-6-astra

Checklist
=========
* [X] Closes #xxx - N/A
* [ ] This pull request requires changes to the ROM set - N/A
  * [ ] I have opened a roms pull request - https://github.com/86Box/roms/pull/changeme/ - N/A
* [ ] This pull request requires changes to the asset set - N/A
  * [ ] I have opened an assets pull request - https://github.com/86Box/assets/pull/changeme/ - N/A
* [ ] This pull request adds, changes or removes an external dependency - N/A
  * [ ] I have opened a docs pull request - https://github.com/86Box/docs/pull/changeme/ - N/A

References
==========
Raw disk images of IBM JX PC DOS 2.10 and Your IBM PC JX Sampler 1.0, available from WinWorldPC.

### DIFF
--- a/src/floppy/fdd_img.c
+++ b/src/floppy/fdd_img.c
@@ -50,6 +50,7 @@ typedef struct img_t {
     int      xdf_type; /* 0 = not XDF, 1-5 = one of the five XDF types */
     int      dmf;
     int      track;
+    int      physical_track;
     int      track_width;
     uint32_t base;
     uint8_t  gap2_size;
@@ -425,6 +426,9 @@ write_back(int drive)
     int    ssize = 128 << ((int) dev->sector_size);
     int    size;
 
+    if ((dev->track < 0) || (dev->track >= dev->tracks))
+        return;
+
     if (dev->is_ioctl) {
         for (int side = 0; side < dev->sides; side++) {
             for (int sector = 0; sector < dev->sectors; sector++) {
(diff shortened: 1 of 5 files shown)


JoshRodd:
Merging problems fixed, which were all in the tests and attendant CMake anyway.

PR #7940, opened 2026-09-12
MIDI: Fix input queue overflow hang and broken real-time bytes

### LINKED ISSUE #6266: MIDI Input Causes Crash
### Checklist

- [x] I **have searched** the issue tracker and **was unable** to find an [open](../issues?q=is%3Aissue+is%3Aopen) or [closed](../issues?q=is%3Aissue+is%3Aclosed) issue matching what I'm seeing.
- [x] I **have verified** that the issue is happening in the **[latest nightly build](https://ci.86box.net/job/86Box/lastSuccessfulBuild/artifact/)**, or the latest commit, if building from source.

### What happened?

MIDI Input crashes 86Box in a variety of VM configurations booting DOS 6.22 and running a variety of MIDI programs that accept MIDI In. Crash occurs when using the standalone MPU-401 emulation with the default address and IRQ, occurs on both Windows and MacOS, and when using different sources of input, from real MIDI controllers and interfaces to virtual ports. Also with both new and old dynarec.

Steps to reproduce: Create a new VM for MS-DOS 6.22, enable MIDI In device and select a MIDI controller, enable "Standalone MPU-401" and use default IRQ and address. Once booted to DOS, load any program that accepts MIDI In, whether a MIDI monitor or sequencer, and play a series of random notes on the controller.

A quick and easy MIDI monitor to use is on the MidiMan MM-401 Floppy Disk  floppy image available from Dos Days here: https://www.dosdays.co.uk/media/midi/MidiMan/midiman_mm-401.rar

Insert the disk and run using "viewer 330" at the A: prompt. It can vary in how many notes it takes to reproduce. Sometimes the MIDI input simply ceases without crashing, but more often than not the VM crashes out with a message saying "Exit Code 11"

I'm happy to provide whatever extra information you may need, and happy to provide my specific VM files if that would be helpful.

### Configuration file

```ini
[MIDI Input Device]
midi_input = 1
realtime = 0
thruchan = 0
clockout = 0

[Roland MPU-IPC-T]
base = 0330
irq = 2
receive_input = 1

[General]
emu_build_num = 7894
force_43 = 1
host_cpu = Apple M4
uuid = bfdca86d-0492-5c61-808f-c604095e9041
vid_renderer = qt_software

[Machine]
cpu_family = celeron_mendocino
cpu_multi = 6
cpu_speed = 400000000
cpu_use_dynarec = 1
fpu_type = internal
machine = p6bat
mem_size = 65536

[Video]
gfxcard = vga

[Input devices]
keyboard_type = keyboard_at
mouse_type = none

[Sound]
midi_in_device = midi_in
mpu401_standalone = 1

[Network]
net_01_link = 0
net_02_link = 0
net_03_link = 0
net_04_link = 0

[Hard disks]
hdd_01_fn = 486dx266.vhd
hdd_01_ide_channel = 0:0
hdd_01_parameters = 63, 16, 1024, 0, ide
hdd_01_speed = ramdisk
hdd_01_vhd_blocksize = 4096

[Floppy and CD-ROM drives]
fdd_01_image_history_01 = MM-401.img
fdd_01_image_history_02 = MM-401.img
fdd_01_type = 35_2hd
fdd_02_type = none
```

### Operating system

MacOS 15.6.1

### 86Box version

v5.2 build 7894

### Build architecture

macOS - Universal (Intel and Apple Silicon)

### Build type

- [ ] New recompiler
- [ ] Debug build

### Additional context

_No response_

MaxFreedomPollard:
Summary
=======

Two defects in the MPU-401 MIDI input path. One can hang the emulator, the other sends the wrong bytes to the MIDI output port and can crash 86Box when no output device is open.

MPU401_RecQueueBuffer (src/sound/snd_mpu401.c:196) never returns once the record queue is full: `cnt` is only incremented inside `if (mpu->rec_queue_used < MPU401_INPUT_QUEUE)`, so when that test fails `cnt < len` can never become false. Only the guest reading the data port (:769) and the block at the end of this same function drain the queue, and the spin reaches neither. MPU401_ClrQueue (:235) and the SysEx abort path (:1391) do zero `rec_queue_used`, but two call sites are reachable from the emulation thread (:482 and :490, in the command write handler), where the spinning thread is the one that would have to issue that reset.

Filling the queue does not need an unusual guest. In intelligent mode any real-time byte above 0xF8 is queued as two bytes with no room check (:1555-1560), and RtMidi is deliberately configured not to ignore active sensing (src/sound/midi_rtmidi.cpp:188). With the guest not reading the port, a controller sending active sensing every 300 ms fills the 1024-byte queue on message 512, about two and a half minutes, and the call that queues the next message never returns.

The loop now stops when there is no room and drops the rest of the event, which is what MPU401_InputSysex already does when the queue is full (:1401-1409), and dropping is the only alternative to blocking here. InputSysex itself is unchanged, since it already clamps to the free space. This does not conflict with the MIDI input queuing work mentioned in #6266: the loop is unbounded whatever the threading, and a rework of the input queue would replace it along with the queue.

The second fix is in midi_raw_out_rt_byte (src/sound/midi.c:310). It checks `midi_in`, then calls play_msg, which dereferences `midi_out` with no check (midi.c:219). `midi_out` is NULL whenever MIDI Out is None or the host port failed to open, and the settings dialog allows an MPU-401 with MIDI In and no MIDI Out. The configuration attached to #6266 is exactly that, `midi_in_device = midi_in` under `[Sound]` and no `midi_device` key.

The formatting is wrong too, whenever the device's Real time option is on. `val << 24` puts the status byte at offset 3 of the int on a little-endian host, so rtmidi_play_msg reads `msg[0]` as zero, takes `midi_lengths[0]` which is 3, and sends three zero bytes instead of the single F8, FA, FB or FC (src/sound/midi_rtmidi.cpp:51 and :59-64). Anything slaved to the emulated MPU-401 clock never starts, stops or syncs. midi_raw_out_byte (midi.c:334) already gets both parts right, so this now guards `midi_out` and uses `midi_out->midi_rt_buf` the same way, leaving the MIDI In policy checks above it alone. That removes the only write to `midi_t.midi_cmd_r` (src/include/86box/midi.h:61), which was already write-only. I left the field in place to keep the diff to the bug, happy to drop it if you would rather.

Checked with two standalone programs rather than a full build, since Qt is not available here. The first runs the fill loop verbatim with the real MPU401_INPUT_QUEUE and the real call shape from the active sensing path, nothing draining the queue, with an iteration cap so it reports instead of hanging: current code reaches `rec_queue_used=1024` on message 512 and does not terminate on 513, the patched loop returns from every call. The second prints what play_msg receives, `00 00 00 f8` at length 3 before and `f8 00 00 00` at length 1 after. Both files compile with `clang -fsyntax-only -Wall -Wextra` with exactly the same warnings as before, and the changed lines are clang-format clean.

Neither fix closes #6266. The IRQ 2 crash there is the ACPI SCI redirect already identified in the thread, and that report has `realtime = 0`. The overflow spin is a plausible cause of the "MIDI input simply ceases" behaviour described in the same report, but I have not reproduced that here.

Checklist
=========

References
==========

* System Real-Time messages are single bytes, 0xF8 to 0xFF: https://midi.org/summary-of-midi-1-0-messages
* Related report, not closed by this pull request: https://github.com/86Box/86Box/issues/6266

### DIFF
--- a/src/sound/midi.c
+++ b/src/sound/midi.c
@@ -318,9 +318,11 @@ midi_raw_out_rt_byte(uint8_t val)
     if (!midi_in->midi_clockout && (val == 0xf8))
         return;
 
-    midi_in->midi_cmd_r = val << 24;
-    /* pclog("Play RT Byte msg\n"); */
-    play_msg((uint8_t *) &midi_in->midi_cmd_r);
+    if (!midi_out || !midi_out->m_out_device)
+        return;
+
+    midi_out->midi_rt_buf[0] = val;
+    play_msg(midi_out->midi_rt_buf);
 }
 
 void
--- a/src/sound/snd_mpu401.c
+++ b/src/sound/snd_mpu401.c
@@ -210,7 +210,8 @@ MPU401_RecQueueBuffer(mpu_t *mpu, uint8_t *buf, unsigned int len)
                 break;
             }
             cnt++;
-        }
+        } else
+            break; /* Input queue full, drop the rest of the message. */
     }
     if (!mpu->queue_used) {
         if (mpu->state.rec_copy || mpu->state.irq_pending) {


OBattler:
Merged, thanks! Also, the IRQ 2 problems are due to some machines' BIOS'es not doing the IRQ 9 to 2 redirect correctly and now that you made me think, it may be due to the machine assigning IRQ 9 to something else. So that's not a bug on our side.

PR #7953, opened 2026-09-14
device: allow a non-ACPI soft power button

JoshRodd:
Summary
=======
The current soft power button is ACPI only. This adds a non-ACPI option for non-ACPI machines, such as the IBM PC Convertible.

This doesn't make any changes to any existing machines, but is necessary for future IBM PC Convertible (model 5140) support.

Checklist
=========
* [X] Closes #xxx - N/A
* [ ] This pull request requires changes to the ROM set - N/A
  * [ ] I have opened a roms pull request - https://github.com/86Box/roms/pull/changeme/ - N/A
* [ ] This pull request requires changes to the asset set - N/A
  * [ ] I have opened an assets pull request - https://github.com/86Box/assets/pull/changeme/ - N/A
* [ ] This pull request adds, changes or removes an external dependency - N/A
  * [ ] I have opened a docs pull request - https://github.com/86Box/docs/pull/changeme/ - N/A

References
==========
None used.


src/qt/qt_mainwindow.ui
```diff
@@ -98,6 +98,7 @@
     <addaction name="actionNon_maskable_interrupt"/>
     <addaction name="separator"/>
     <addaction name="actionACPI_Shutdown"/>
+    <addaction name="actionHard_Power_Loss"/>
```
lemondrops: Why not just rename and repurpose the existing button?

lemondrops (changes requested):

lemondrops:
I don't see the point of multiple separate "power off" options: just one generic "soft power off" button that does ACPI/APM/PC Convertible/whatever method depending on the machine should be enough, and "hard power off" is equivalent to exiting 86Box.

src/qt/qt_mainwindow.cpp
```diff
@@ -1853,9 +1853,14 @@ MainWindow::refreshMediaMenu()
     status->setDynarecMenu(dynarecMenu);
     status->refresh(ui->statusbar);
     ui->actionMCA_devices->setVisible(machine_has_bus(machine, MACHINE_BUS_MCA));
+    const bool has_power_button = device_has_power_button();
+    ui->actionForce_4_3_display_ratio->setEnabled(monitors[0].mon_device_aspect_x == 0);
```
lemondrops: `mon_device_aspect_x` doesn't exist, also how is this related to the PR in question?

src/qt/qt_mainwindow.cpp
```diff
@@ -1853,9 +1853,14 @@ MainWindow::refreshMediaMenu()
...
     if (acpi_enabled) {
         ui->actionACPI_Shutdown->setText(tr("ACP&I shutdown"));
         ui->actionACPI_Shutdown->setToolTip(tr("ACPI shutdown"));
+    } else if (has_power_button) {
+        ui->actionACPI_Shutdown->setText(tr("Non-ACP&I soft power button"));
+        ui->actionACPI_Shutdown->setToolTip(tr("Non-ACPI soft power button"));
```
lemondrops: I recommend renaming this to something more generic instead of changing the name on-the-fly.

lemondrops (changes requested):

src/qt/qt_mainwindow.cpp
JoshRodd: Agreed - I'll just go with "Power button (soft)" for now; we won't mention ACPI anymore. As long as this won't confuse people looking for the old "ACPI power button" option.

src/qt/qt_mainwindow.cpp
JoshRodd: It wasn't. This sloppily leaked out of my unfinished PC Convertible code (that machine has a non-4:3 monitor.)

lemondrops (approved):

PR #7984, opened 2026-09-18
cpu: follow the IRET/IRETD EFLAGS rules in pmodeiret()

gerben-stavenga:
pmodeiret() applies no privilege check to the flags the SDM's IRET/IRETD
operation restricts, in two places:

  - the V86-entry branch is taken on the frame's VM bit alone, where the
    operation reads

        IF tempEFLAGS(VM) = 1 and CPL = 0
            THEN GOTO RETURN-TO-VIRTUAL-8086-MODE;

  - the 32-bit path then restores the entire high EFLAGS word from the
    frame, "cpu_state.eflags = tempflags >> 16", where
    RETURN-TO-{SAME,OUTER}-PRIVILEGE-LEVEL read

        IF OperandSize = 32 or 64  THEN EFLAGS(RF, AC, ID) := tempEFLAGS;
        IF CPL = 0
            THEN EFLAGS(IOPL) := tempEFLAGS;
                 IF OperandSize = 32 or 64
                     THEN EFLAGS(VIF, VIP) := tempEFLAGS;

    VM is absent from both lists: a protected-mode return never loads it.

So gate the V86 path on CPL 0, and build a mask for the high word out of
exactly those bit sets, from the CPL of the IRET itself, read before CS
is reloaded. IOPL and IF already have their rules applied by flagmask
directly above.

Two observable changes. A ring-3 IRETD can no longer switch the machine
into V86 mode: with VM set in its frame it now pops three dwords and
returns to protected mode, instead of popping nine and reprogramming
CS/SS/ES/DS/FS/GS through do_seg_v86_init(). And a protected-mode
virtual interrupt (CR4.PVI) that the emulator recorded in VIF while a
ring-3 client had interrupts virtually disabled is no longer dropped
when that client IRETDs out of its own handler with a frame captured
while VIF was clear; previously the client waited forever for an
interrupt it had already been handed.

The VIF case turned up under the RetroOS DPMI host
(https://github.com/gerben-stavenga/RetroOS), whose strict-PVI clients,
Dark Forces among them, hang with VIF=0 on current master. Its DOS
personality enters V86 mode by IRETD and runs DPMI clients at ring 3,
so it exercises both paths; it boots and passes its DOS/DPMI and SB16
regression suite with this change.

cpu_state.eflags only ever holds RF, VM, AC, VIF, VIP and ID, so the
mask drops no emulated state; it also stops the reserved EFLAGS bits
22-31 being stored there. AC_FLAG gets a name in cpu.h alongside its
neighbours, as POPFD spells it 0x04 by hand today.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


JoshRodd:
Hi @gerben-stavenga and thank you for your contribution!

Can you tell us if you have confirmed that your software works correctly on a real 386 or other real x86 hardware?

In addition, in the future, please follow the contributor guide at https://github.com/86Box/86Box/blob/master/CONTRIBUTING.md with regards to using AI assistance:

> The author must understand what it actually does and while using AI to assist with coding is fine, pull requests written solely by AI are more likely than not to be rejected

gerben-stavenga:
> Hi @gerben-stavenga and thank you for your contribution!
> 
> Can you tell us if you have confirmed that your software works correctly on a real 386 or other real x86 hardware?
> 
> In addition, in the future, please follow the contributor guide at https://github.com/86Box/86Box/blob/master/CONTRIBUTING.md with regards to using AI assistance:
> 
> > The author must understand what it actually does and while using AI to assist with coding is fine, pull requests written solely by AI are more likely than not to be rejected

Hi, I added PVI support to my retro operating system (https://github.com/gerben-stavenga/RetroOS) in order to mitigate CLI/STI trapping overhead when running old dos games. It worked properly on qemu (tcg and importantly kvm, which is runs natively) and when I run my OS natively on my AMD. But on 86box games hang on VIF=0 loops, luckily, nowadays, I had a clanker trace the root cause to this and verified it works properly now. I do not have access to 383/486 or pentium, only my modern amd, this change seemed to me correct with respect to intel spec (and very small to audit/review), so I hoped it would be welcome despite root-caused by AI.

JoshRodd:
> Hi, I added PVI support to my retro operating system (https://github.com/gerben-stavenga/RetroOS) in order to mitigate CLI/STI trapping overhead when running old dos games. It worked properly on qemu (tcg and importantly kvm, which is runs natively) and when I run my OS natively on my AMD. But on 86box games hang on VIF=0 loops, luckily, nowadays, I had a clanker trace the root cause to this and verified it works properly now. I do not have access to 383/486 or pentium, only my modern amd, this change seemed to me correct with respect to intel spec (and very small to audit/review), so I hoped it would be welcome despite root-caused by AI.

Hi, access to any modern x86 CPU would do (they can all boot DOS, etc., and should be able to boot your "retro" operating system as well). Can you please confirm that a modern AMD CPU operates in the same way as your change claimed - perhaps making a very small testbed that can easily be booted (you could package it as an EFI executable along with an old fashioned boot sector or .DOS program) so we can test this out? I am having trouble figuring out how to reproduce this.

The reason I ask is because many operating systems that do all kinds of things with the interrupt flag, including strange, undocumented ways of using it, like OS/2 3.0 and 4.0, have not had any problems with 86Box, and we want to ensure we are not introducing a behaviour that is not reflected in real-world CPUs.

Additionally, I'm hoping you can agree with 86Box's contributor stance to hand-write PRs - I also use extensive AI assistance but we want to ensure "humans are in the loop" and that we don't have AI agents autonomously writing and submitting PRs. Thoughts?

gerben-stavenga:
> > Hi, I added PVI support to my retro operating system (https://github.com/gerben-stavenga/RetroOS) in order to mitigate CLI/STI trapping overhead when running old dos games. It worked properly on qemu (tcg and importantly kvm, which is runs natively) and when I run my OS natively on my AMD. But on 86box games hang on VIF=0 loops, luckily, nowadays, I had a clanker trace the root cause to this and verified it works properly now. I do not have access to 383/486 or pentium, only my modern amd, this change seemed to me correct with respect to intel spec (and very small to audit/review), so I hoped it would be welcome despite root-caused by AI.
> 
> Hi, access to any modern x86 CPU would do (they can all boot DOS, etc., and should be able to boot your "retro" operating system as well). Can you please confirm that a modern AMD CPU operates in the same way as your change claimed?
> 

So ran my OS and it now behaves the same when run under 86box as well as when i run it native, before this os would misbehave on 86box. 

> The reason I ask is because many operating systems that do all kinds of things with the interrupt flag, including strange, undocumented ways of using it, like OS/2 3.0 and 4.0, have not had any problems with 86Box, and we want to ensure we are not introducing a behaviour that is not reflected in real-world CPUs.
> 
> Additionally, I'm hoping you can agree with 86Box's contributor stance to hand-write PRs - I also use extensive AI assistance but we want to ensure "humans are in the loop" and that we don't have AI agents autonomously writing and submitting PRs. Thoughts?

I understand, you can be assured I was in the loop. Mostly I used AI to write the description with the relevant snippets from felixcloulier instruction spec website. What I will do over the weekend is add a simple program that exercises this path and reads the flags and run it native and within 86box. And post it  as "out-of-the-box" test.

gerben-stavenga:
I added a test https://github.com/gerben-stavenga/RetroOS/releases/download/vifiret-hdpmi/vifiret-hdpmi-freedos.zip

i used hdpmi32i which runs apps in ring-3 with setpvi to enable PVI in freedos. 
qemu with kvm - test pass (this executes instructions natively)
old 86box - fail
with this pr - pass

PR #8012, opened 2026-09-20
cdrom: implement CDROM_BUS_LPT, and add the Micro Solutions BackPack bridge

Mike1978uk:
`CDROM_BUS_LPT` has been declared in `cdrom.h` and implemented by nothing, so a CD-ROM could not be attached to a parallel-port bridge. This implements it and adds a bridge that uses it.

`scsi_cdrom.c` gains an LPT arm beside the SCSI one, and a `cdrom_get_lpt_device()` so a bridge can fetch the drive assigned to its port. `cdrom.c` resets those drives; `config.c` reads `cdrom_XX_lpt_port`. Configure with `cdrom_01_parameters = 1, lpt`.

**One fix worth noting on its own:** `scsi_cdrom_current_mode()` returned 2 for SCSI, 1 for ATAPI and 0 for anything else, and 0 means "no transfer". Any bus reaching that default gets a drive that enumerates but returns no data for any command, INQUIRY included. A parallel bridge is PIO, and now says so.

`lpt_epat.c` now falls back to a CD-ROM when no removable disk is on its port — the EPAT is a generic parallel-to-ATAPI adapter and was sold carrying both.

`lpt_bpck.c` models the Micro Solutions BackPack CD-ROM: the wire protocol, the 93C46 identity EEPROM, and an ATAPI task file at register `0x40` served by 86Box's own CD-ROM.

**Tested** against a real Micro Solutions BackPack Bantam CD-ROM (model 180100, Toshiba XM-1502B mechanism). Every layer was measured on that drive over the parallel port rather than inferred: the chain address, the connect handshake, the presence test, the register framing, and the nibble-decode table, which was taken from the vendor driver and checked entry by entry. The identity EEPROM was read off the drive itself with a small DOS program and is reproduced verbatim. None of this could have been derived from the driver binary alone — the hardware settled several points where reasoning from the disassembly gave the wrong answer.

In 86Box, the vendor DOS driver **in its own installer's configuration** enumerates the drive and MSCDEX reads an ISO.

**Not tested:** EPP. The driver also attempts PS/2 byte mode and this device answers in it, but the driver falls back to nibble mode; that may be the emulated parallel port not offering a bidirectional data path rather than anything in this device — it has not been investigated. The EPAT-carrying-a-CD path is compiled and reasoned about, not exercised — no EPAT CD-ROM driver was available. One drive only; other BackPack variants are unknown.

Based on #8010, which adds the `lpt` bus keyword this needs.


Mike1978uk:
Minor wording correction pushed: comments in `lpt_bpck.c` still referred to the LS-120 engine it was ported from, and two dated diagnostics came along with them. Comments only, no functional change, retested.

jriwanek:
src/char/char.c has merge conflicts which need to be resolved.

Mike1978uk:
Same rebase — `src/char/char.c` resolved and both branches are on `e08ee5a52`. Details on #8010. Thanks.

src/char/char.c
```diff
@@ -65,6 +65,8 @@ static const struct {
     { &lpt_prt_ps_device },
     { &lpt_prt_pcl_device },
     { &lpt_hasp_savquest_device },
+    { &lpt_bpck_device },
+    { &lpt_epat_device },
```
dhrdlicka: I don't think this belongs here

Mike1978uk: Changed in both, thanks — you're right, and it now follows `lpt_ditto` since `e684a9add`.

Both bridges are instantiated from the drive's own bus assignment instead of the dropdown:

```c
if (rdisk_drives[c].bus_type == RDISK_BUS_LPT)      /* #8010, rdisk_hard_reset */
    device_add_inst(&lpt_epat_device, rdisk_drives[c].res + 1);

case CDROM_BUS_LPT:                                 /* #8012, cdrom_hard_reset */
    device_add_inst(&lpt_bpck_device, dev->res + 1);
```

`src/char/char.c` is untouched in both PRs now — it has dropped out of the diffs entirely.

Tested both routes rather than just rebuilt, each with **no `lpt1_device` line in the config at all**, so the bridge can only have come from the drive assignment. On #8010 the LS-120 enumerates, takes a drive letter, its directory reads and a 92,870-byte file lands — free space 124,841,984 → 124,747,776. On #8012 the vendor DOS driver enumerates the BackPack and MSCDEX lists the disc. Both identical to the results before the change.

One thing removed rather than worked around: the commit letting the EPAT carry a CD-ROM. It was never exercised — there's no EPAT CD-ROM driver here to test with — and once the bridge is implied by bus type it would have been unreachable anyway.

What these two PRs cover is the LS-120 and the BackPack, both built against real hardware. Another parallel device would need its own design, model and testing rather than a flag on these — but the foundation is in place now for anyone who wants to add one: `RDISK_BUS_LPT` and `CDROM_BUS_LPT` are implemented, the ECP callbacks exist on `lpt_device_t`, and there are two worked examples of a bridge to follow.

